PR #13, opened 2021-12-16
configurable options per facet

yen-tt:
To match the customization level in SDK, this pr update Facets and Facet component to allow configuration of individual facet for following options: searchable, collapsible, defaultExpanded, label, and placeholderText. If not provided, Facet component will use the top-level configuration for Facets and any default value in place.

J=SLAP-1764
TEST=manual

add facetConfigs to Facets in LocationPage with different configuration combo, and see that it works as expected.

### DIFF
--- a/src/components/Facet.tsx
+++ b/src/components/Facet.tsx
@@ -5,16 +5,16 @@ import { CompositionMethod, useComposedCssClasses } from '../hooks/useComposedCs
 
 export type onFacetChangeFn = (fieldId: string, option: DisplayableFacetOption) => void
 
-export interface FacetTextConfig {
+export interface OverridableFacetConfig {
+  searchable?: boolean,
   placeholderText?: string,
-  label?: string
+  label?: string,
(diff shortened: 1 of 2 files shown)


src/components/Facet.tsx
```diff
@@ -5,16 +5,16 @@ import { CompositionMethod, useComposedCssClasses } from '../hooks/useComposedCs
 
 export type onFacetChangeFn = (fieldId: string, option: DisplayableFacetOption) => void
 
-export interface FacetTextConfig {
+export interface OverridableFacetConfig {
```
cea2aj, line 8: I have a slight preference towards calling this `FacetConfig` for simplicity. What are your thoughts on that?

yen-tt: I just remember a nit comment from before where FacetConfig vs FacetProps may be confusing? Although I'm ok with FacetConfig as well.

cea2aj (approved):

PR #16, opened 2021-12-17
add tailwind styling for DirectAnswer

oshi97:
Did not add feedback buttons, which are planned for a separate item.

J=SLAP-1759
TEST=manual

check that both featured snippet and field value direct answers match the mocks
used the us_presidents experience for featured snippet direct answers
used mock link data, because we don't support using custom fields


src/components/DirectAnswer.tsx
```diff
@@ -1,52 +1,98 @@
...
 }
 
-const defaultCSSClasses = {
-  container: 'DirectAnswer',
-  containerLoading: 'DirectAnswer--loading',
-  title: 'DirectAnswer__title',
-  content: 'DirectAnswer__content',
-  description: 'DirectAnswer__contentDescription'
+const defaultCSSClasses: DirectAnswerCssClasses = {
+  container: 'p-4 border border-color-zinc-200',
```
cea2aj, line 24: Can we update the zinc to a gray? IIRC, the mocks only use gray and blue colors

src/components/DirectAnswer.tsx
```diff
@@ -1,52 +1,98 @@
...
+  containerLoading: 'opacity-50',
```
cea2aj, line 25: Can we use `container___loading` since it's a state?

src/components/DirectAnswer.tsx
cea2aj, line 24: Can we add a border radius and a shadow just like the cards?

cea2aj, line 24: I think we should add some margin because right now it's right next to the nav bar

src/components/DirectAnswer.tsx
```diff
@@ -1,52 +1,98 @@
...
     </div>
   )
 }
+
+function getCssClasses(
+  customCssClasses: DirectAnswerCssClasses | undefined,
+  isLoading: boolean,
+  directAnswerResult: DirectAnswerModel
+) {
+  const cssClasses = Object.assign(defaultCSSClasses, customCssClasses);
```
yen-tt, line 86: I think we can add cssCompositionMethod in DirectAnswerProps, and pass it to here as well to use useComposedCssClasses instead of assign

src/components/DirectAnswer.tsx
```diff
@@ -1,52 +1,98 @@
...
-        <div className={cssClasses.description}>
-          {directAnswerResult.type === DirectAnswerType.FeaturedSnippet 
-            ? renderHighlightedValue(directAnswerResult.snippet)
-            : directAnswerResult.value}
-        </div>
-        {directAnswerResult.relatedResult.link 
-          && <a href={directAnswerResult.relatedResult.link}>View More</a>}
+        <div className={cssClasses.description}>{description}</div>
+        {link && 
+          <div className={cssClasses.viewDetailsLinkContainer}>
```
yen-tt, line 72: I'm thinking we can just directly use whenever return by `getLinkText()` instead of adding an additional div here? and the return element for `if(isSnippet && name)` can be a span/div instead of a div fragment

src/config/answersHeadlessConfig.ts
```diff
@@ -1,4 +1,8 @@
-export const answersHeadlessConfig = {
+import { AnswersHeadlessProvider } from '@yext/answers-headless-react';
+
+type HeadlessProviderProps = Parameters<typeof AnswersHeadlessProvider>[0];
```
yen-tt, line 3: oh nice. but maybe we should consider exporting AnswersHeadlessProviderConfig/Props in answers-headless-react so we don't have to derive it this way?

cea2aj: Makes sense to me. I can make a separate item for that

src/components/DirectAnswer.tsx
cea2aj: updated to remove the zinc

src/components/DirectAnswer.tsx
cea2aj: updated

src/components/DirectAnswer.tsx
cea2aj: updated

cea2aj: updated to give the nav bar bottom margin

src/components/DirectAnswer.tsx
cea2aj: Updated to use `useComposedCssClasses`

src/components/DirectAnswer.tsx
cea2aj: I moved the viewDetailsLinkContainer into `getLinkText` which means the div won't render unless it's needed.

yen-tt (approved):

PR #17, opened 2021-12-20
Alternative verticals tailwind

nmanu1:
Add tailwind styling for `AlternativeVerticals`.

J=SLAP-1758
TEST=manual

Search for 'jobs' in the FAQs vertical and see that the alternative verticals match the Figma mock


cea2aj:
Do you have some example queries which return no results? I'm performing some searches which return no results, but I'm not seeing the Alternative Vertical component

src/components/AlternativeVerticals.tsx
```diff
@@ -85,41 +126,37 @@ export default function AlternativeVerticals (props: Props): JSX.Element | null
...
   }
 
   function renderUniversalDetails() {
     return (
-      <div className='AlternativeVerticals__universalDetails'>
-        <span>Alternatively you can </span>
-        <button className='AlternativeVerticals__button AlternativeVerticals__universalDetailsButton' 
+      <div className={cssClasses.categoriesText}>
+        <span>View results across </span>
+        <button className={cssClasses.allCategoriesLink}
```
yen-tt, line 157: Can we update this to a Link component that redirects to 'All' page?

yen-tt:
> Do you have some example queries which return no results? I'm performing some searches which return no results, but I'm not seeing the Alternative Vertical component

you should see it if you search for 'jobs' in FAQs

src/icons/star.svg
```diff
@@ -1,3 +1,3 @@
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 18 18">
+<svg className="fill-current" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 18 18">
```
cea2aj, line 1: If I wanted to change this color, which value would I change?

cea2aj, line 1: I am a little hesitant to put tailwind in here because this wouldn't work if tailwind weren't on the page. What do you think about setting this with inline styling? E.g. (style="fill: currentColor;")

nmanu1: To change this, you'd change the color in the `suggestions` css class. This would change the color of the text and the icon since Rose said they should be the same

nmanu1: That makes sense, I'll change it

src/components/AlternativeVerticals.tsx
```diff
@@ -85,41 +126,37 @@ export default function AlternativeVerticals (props: Props): JSX.Element | null
...
       </div>
     );
   }
 
   function renderSuggestion(suggestion: VerticalSuggestion) {
     return (
-      <li key={suggestion.verticalKey} className="AlternativeVerticals__suggestion">
-        <button className='AlternativeVerticals__suggestionLink AlternativeVerticals__button'
+      <li key={suggestion.verticalKey} className={cssClasses.suggestion}>
+        <button className={cssClasses.suggestionButton}
```
yen-tt, line 141: and here too, can we change this to Link that redirects to where this suggestion's vertical page should be?

src/components/AlternativeVerticals.tsx
cea2aj: We actually have a separate item for updating the routing of this component: https://yexttest.atlassian.net/browse/SLAP-1766

yen-tt: oh cool

src/components/AlternativeVerticals.tsx
yen-tt: nvm, there's an item for this: https://yexttest.atlassian.net/browse/SLAP-1766

cea2aj (approved):

PR #18, opened 2021-12-22
Vertical State Management Update

cea2aj:
Use the `searchType` provided by answers headless and remove the `isVertical` props from components

J=SLAP-1757
TEST=manual

Build the sample app and smoke test


src/components/SearchBar.tsx
```diff
@@ -67,12 +65,13 @@ export default function SearchBar({
   const answersActions = useAnswersActions();
   const query = useAnswersState(state => state.query.input);
   const isLoading = useAnswersState(state => state.searchStatus.isLoading);
+  const isVertical = useAnswersState(s => s.meta.searchType) === 'vertical';
```
yen-tt, line 68: nit: can we use SearchTypeEnum instead of literal string?

cea2aj: Sure

yen-tt (approved):

PR #19, opened 2021-12-23
Update AlternativeVerticals to use React Router

nmanu1:
Use `Link` to update the URL and navigate to the correct universal/vertical page, instead of showing the results on the current page.

J=SLAP-1766
TEST=manual

Click on the vertical and universal links in the alternative verticals component and check that they navigate to the correct page.

### DIFF
--- a/src/components/AlternativeVerticals.tsx
+++ b/src/components/AlternativeVerticals.tsx
@@ -3,6 +3,7 @@ import { ReactComponent as Star } from '../icons/star.svg';
 import { useAnswersState, useAnswersActions, VerticalResults } from '@yext/answers-headless-react';
 import { CompositionMethod, useComposedCssClasses } from '../hooks/useComposedCssClasses';
 import classNames from 'classnames';
+import { Link } from 'react-router-dom';
 
 interface AlternativeVerticalsCssClasses {
   container?: string,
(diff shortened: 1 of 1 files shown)


src/components/AlternativeVerticals.tsx
```diff
@@ -154,10 +150,9 @@ export default function AlternativeVerticals ({
     return (
       <div className={cssClasses.categoriesText}>
         <span>View results across </span>
-        <button className={cssClasses.allCategoriesLink} 
-          onClick={actions.executeUniversalQuery}>
+        <Link className={cssClasses.allCategoriesLink} to={`/?query=${query}`}>
           <span>all search categories.</span>
```
yen-tt, line 154: is the span tag needed? can we have literal string directly in the Link component instead

src/components/AlternativeVerticals.tsx
```diff
@@ -138,14 +138,10 @@ export default function AlternativeVerticals ({
   function renderSuggestion(suggestion: VerticalSuggestion) {
     return (
       <li key={suggestion.verticalKey} className={cssClasses.suggestion}>
-        <button className={cssClasses.suggestionButton}
-          onClick={() => {
-            actions.setVertical(suggestion.verticalKey);
-            actions.executeVerticalQuery();
-          }}>
+        <Link className={cssClasses.suggestionButton} to={`/${suggestion.verticalKey}?query=${query}`}>
```
yen-tt, line 141: I don't think `?query=${query}` part is needed for the router

src/components/AlternativeVerticals.tsx
nmanu1: you're right, it's not needed so I removed it

src/components/AlternativeVerticals.tsx
nmanu1: removed!

yen-tt (approved):

PR #21, opened 2021-12-23
Visual Autcomplete Styling Updates

cea2aj:
Update the styling of the Visual Autocomplete Search Bar.

Align the recent searches with the regular autocomplete results. Add a divider above the FAQ results. Adjust positioning of the vertical links.

J=SLAP-1761
TEST=visual

Visually inspect the dropdown with some mocked and real data


src/components/utils/renderAutocompleteResult.tsx
```diff
@@ -7,8 +7,8 @@ export interface AutocompleteResultCssClasses {
 }
 
 export const builtInCssClasses = {
-  option: 'flex whitespace-pre-wrap',
-  icon: 'w-8 mx-2'
+  option: 'flex whitespace-pre-wrap h-6.5 pl-3',
+  icon: 'w-6 text-gray-400 opacity-50'
```
tmeyer2115, line 11: This will also change the styling for the non-VA query suggestions. Is that alright?

src/components/VisualAutocomplete/SampleVisualSearchBar.tsx
```diff
@@ -2,6 +2,7 @@ import VisualSearchBar from './VisualSearchBar';
 import { Result } from '@yext/answers-headless-react';
 import EntityPreviews from './EntityPreviews';
 import { universalResultsConfig } from '../../config/universalResultsConfig';
+import { Divider } from '../../components/StaticFilters';
```
tmeyer2115, line 5: Nit: It's a bit odd for the VA Component to have a dependency on the `StaticFilters` file. Maybe the `Divider` should be moved to a common location.

src/components/VisualAutocomplete/SampleVisualSearchBar.tsx
```diff
@@ -47,12 +48,16 @@ interface FaqData {
 
 function FaqCard({ result }: CardProps) {
   const faqData: FaqData = result.rawData;
-  return (
-    <div tabIndex={0} className='flex flex-col m-4 rounded-md p-2'>
-      <div className='text-lg font-medium pb-1'>{faqData.question}</div>
+  return (<div key={faqData.question}>
+    <Divider
```
tmeyer2115, line 52: It feels a bit odd to have a `Divider` Component. It's simply a `div` and some styling. If we replace the styling here and have wholly different styling in the `StaticFilter` Component, what benefit does the `Divider` really give us?

src/components/utils/renderAutocompleteResult.tsx
cea2aj: Yep, that's alright. I'm updating the SVG to use the "currentColor" fill so that the color of the icon can be controlled here with tailwind

src/components/VisualAutocomplete/SampleVisualSearchBar.tsx
cea2aj: Since we are replacing all the styling, I'll remove the use of the Divider component and render the div directly

src/components/VisualAutocomplete/SampleVisualSearchBar.tsx
cea2aj: I'm removing the dependency on the Divider component, however I think we should decide on how general we want the divider to be. Outside of this file, it's used by Facets, StaticFilters, and the Locations Page to put a divider between filter search and facets. If we want a generic divider, then it's own file makes sense. However if it's simply shared by filters then I think it's okay to keep it in static filters. Or maybe we call it `FilterDivider` so that it's more scoped

tmeyer2115: If it's only used in regard to filters, calling it `FilterDivider` and leaving it where it is sounds alright.

tmeyer2115 (approved):

PR #23, opened 2021-12-23
Navigation dropdown tailwind

nmanu1:
Update the tailwind styling for the navigation dropdown menu.

J=SLAP-1770
TEST=manual

Check that the dropdown menu matches the mocks. When the current vertical is in the menu, the text is blue, instead of gray.


src/components/Navigation.tsx
```diff
@@ -13,17 +14,23 @@ interface NavigationCssClasses {
   menuButton?: string,
   menuButtonContainer?: string,
   menuButton___menuOpen?: string,
-  menuContainer?: string
+  menuButton___hasActiveLink?: string,
+  menuContainer?: string,
+  menuNavLink?: string,
+  menuActiveNavLink?: string
```
cea2aj, line 20: Since active is a state, what do you think of `menuNavLink___active` instead?

src/components/Navigation.tsx
```diff
@@ -82,11 +91,22 @@ export default function Navigation({ links, customCssClasses, cssCompositionMeth
   const { search } = useLocation();
   const visibleLinks = links.slice(0, links.length - numOverflowLinks);
   const overflowLinks = links.slice(-numOverflowLinks);
-  const menuButtonClassNames = cssClasses.menuButton___menuOpen
+  let isActiveLinkInMenu = false;
```
cea2aj, line 94: nit: What do you think of `menuContainsActiveLink`?

src/components/Navigation.tsx
nmanu1: That makes sense. I was basing the naming off of the `navLink` and `activeNavLink` classes (lines 12-13). I can change `activeNavLink` to `navLink___active` to match

src/components/Navigation.tsx
```diff
@@ -82,11 +91,22 @@ export default function Navigation({ links, customCssClasses, cssCompositionMeth
   const { search } = useLocation();
   const visibleLinks = links.slice(0, links.length - numOverflowLinks);
   const overflowLinks = links.slice(-numOverflowLinks);
-  const menuButtonClassNames = cssClasses.menuButton___menuOpen
+  let menuContainsActiveLink = false;
+  overflowLinks.forEach(({ to }) => {
```
oshi97, line 95: i might use array.prototype.find here so it short circuits

src/components/Navigation.tsx
```diff
@@ -82,11 +91,22 @@ export default function Navigation({ links, customCssClasses, cssCompositionMeth
...
+    if (to === currentVertical || (to === '/' && currentVertical === undefined)) {
+      menuContainsActiveLink = true;
+    }
+  })
+  let menuButtonClassNames = cssClasses.menuButton___menuOpen
```
oshi97, line 100: could this be done in one usage of classNames? all the ternary operators make things feel tangled together, at least for me

nmanu1: I couldn't find an easy way to do this with one use of classNames because both `menuButton___menuOpen` and `menuButton___hasActiveLink` are optional classes. So we have to check that each exists before passing it to classNames, but we still would want to apply one even if the other one is not defined. The only more readable way I could think of was something like the following:
```
const menuButtonClassNames = classNames(cssClasses.menuButton, {
  [cssClasses.menuButton___menuOpen ?? '']: menuOpen,
  [cssClasses.menuButton___hasActiveLink ?? '']: menuContainsActiveLink
});
```
But, I was avoiding this based on a previous PR [comment](https://github.com/yext/answers-headless-react/pull/65#discussion_r760331606)

src/components/Navigation.tsx
nmanu1: good point! switched to use array.prototype.some since we don't need the extra info returned by find

src/components/Navigation.tsx
oshi97: does 
```ts
const menuButtonClassNames = classNames(cssClasses.menuButton, {
  [cssClasses.menuButton___menuOpen]: cssClasses.menuButton___menuOpen && menuOpen,
  [cssClasses.menuButton___hasActiveLink]: cssClasses.menuButton___hasActiveLink && menuContainsActiveLink
});
```
work?
it's not perfect but might be a small improvement?

oshi97: I think I prefer your first idea more to be honest as a general pattern but @tmeyer2115 what do you think?

nmanu1: > does
> 
> ```ts
> const menuButtonClassNames = classNames(cssClasses.menuButton, {
>   [cssClasses.menuButton___menuOpen]: cssClasses.menuButton___menuOpen && menuOpen,
>   [cssClasses.menuButton___hasActiveLink]: cssClasses.menuButton___hasActiveLink && menuContainsActiveLink
> });
> ```
> 
> work? it's not perfect but might be a small improvement?

This wouldn't work because the problem is that `cssClasses.menuButton___menuOpen` can't be used as a property if it could be undefined

> I think I prefer your idea of `[cssClasses.menuButton___menuOpen ?? '']` more to be honest, @tmeyer2115 what do you think?

I think I'd also prefer this for readability

tmeyer2115: @nmanu1, I'm fine with this suggestion: 

```
const menuButtonClassNames = classNames(cssClasses.menuButton, {
  [cssClasses.menuButton___menuOpen ?? '']: menuOpen,
  [cssClasses.menuButton___hasActiveLink ?? '']: menuContainsActiveLink
});
```

The `?? ''` is a bit odd, but I think the increased readability and conciseness is well worth it.

tmeyer2115 (approved):

PR #24, opened 2021-12-28
Refactor display checkbox for filter/facet option

yen-tt:
- Created a util function renderCheckboxOption to avoid duplicate code and styling in StaticFilters and Facet(s) component.
- added dropdown icon for Facet component

J=SLAP-1773
TEST=manual

- start sample app, and see that the styling for static filters and facets stay the same. Toggle checkbox and see that it behave as expected
- toggle expand/collapse state from Facet component, see that icon's rotation update accordingly.
- when facet is not collapsible, see that dropdownIcon doesn't appear

### DIFF
--- a/src/components/Facet.tsx
+++ b/src/components/Facet.tsx
@@ -84,7 +84,8 @@ export default function Facet(props: FacetProps): JSX.Element {
           {facetOptions.map(option => 
             renderCheckboxOption({
               option: { id: option.displayName, label: `${option.displayName} (${option.count})` },
-              optionHandler: () => onToggle(facet.fieldId, option)
+              optionHandler: () => onToggle(facet.fieldId, option),
+              selected: option.selected
             })
           )}
         </div>
--- a/src/components/Facet.tsx
+++ b/src/components/Facet.tsx
@@ -2,6 +2,7 @@ import { useAnswersUtilities, DisplayableFacet, DisplayableFacetOption } from '@
 import { useState } from 'react';
 import useCollapse from 'react-collapsed';
 import { CompositionMethod, useComposedCssClasses } from '../hooks/useComposedCssClasses';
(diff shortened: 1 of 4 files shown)


src/components/Facet.tsx
```diff
@@ -53,18 +53,24 @@ export default function Facet(props: FacetProps): JSX.Element {
...
     ? answersUtilities.searchThroughFacet(facet, filterValue).options
     : facet.options;
 
   return (
     <fieldset>
-      <button className={cssClasses.facetLabel} {...(collapsible ? getToggleProps() : {})}>
-        {label || facet.displayName} 
+      <button className={cssClasses.labelContainer} {...(collapsible ? getToggleProps() : {})}>
+        <div className={cssClasses.label}>{label || facet.displayName}</div>
+        <DropdownIcon className={modifiedLabelIconCssClasses}/>
```
tmeyer2115, line 73: Should we always be displaying the drop-down icon? What if the Facet is not collapsible?

src/components/utils/renderCheckboxOption.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+
+const builtInCssClasses: CheckboxOptionCssClasses = {
+  option: 'flex items-center space-x-3',
+  optionInput: 'w-3.5 h-3.5 form-checkbox border border-gray-300 rounded-sm text-blue-600 focus:ring-blue-500',
+  optionLabel: 'text-gray-600 text-sm font-normal'
+}
+
+export default function renderCheckboxOption({
+  option, selected, optionHandler, customCssClasses
```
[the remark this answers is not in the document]
yen-tt: onClick sounds good

src/components/Facet.tsx
yen-tt: ah good point, I will update

tmeyer2115 (approved):

PR #29, opened 2022-01-07
Remove screenReaderInstructionsId prop

nmanu1:
Removes the `screenReaderInstructionsId` prop from `SearchBar` and `InputDropdown`. Instead, `InputDropdown` generates a UUID that is used internally and passed to `ScreenReader`.

Also, fixes a bug in `InputDropdown` that prevented some screen reader announcements from being made and sometimes caused a lag in announcements.

J=SLAP-1780
TEST=manual

Checked that screen reader instructions are still announced correctly and that announcements are now correct for `SearchBar` and `FilterSearch`.

Note: Screen reader announcements for visual autocomplete will be addressed in another item.

### DIFF
--- a/src/components/InputDropdown.tsx
+++ b/src/components/InputDropdown.tsx
@@ -1,5 +1,5 @@
 import classNames from "classnames";
-import React, { useReducer, KeyboardEvent, useRef, useEffect, useState, FocusEvent, Children } from "react"
+import React, { useReducer, KeyboardEvent, useRef, useEffect, useState, useMemo, FocusEvent, Children } from "react"
 import DropdownSection, { DropdownSectionProps } from "./DropdownSection";
 import ScreenReader from "./ScreenReader";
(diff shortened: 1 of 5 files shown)


src/components/InputDropdown.tsx
```diff
@@ -86,11 +85,12 @@ export default function InputDropdown({
   const [latestUserInput, setLatestUserInput] = useState(inputValue);
   const [childrenKey, setChildrenKey] = useState(0);
   const [screenReaderKey, setScreenReaderKey] = useState(0);
+  const screenReaderInstructionsId = uuid();
```
cea2aj, line 88: I believe this approach will give us a new UUID each render which might impact performance. Can we wrap it in `useMemo` so that we don't compute a new UUID each time?

nmanu1: Updated!

cea2aj (approved):

PR #30, opened 2022-01-07
Add FilterDisplayManager

cea2aj:
Add `FilterDisplayManager` for showing and hiding the filters based on the screen width and the view/hide filter buttons

A `PageView` is added to context through the `PageViewContextProvider`, and any component or page can access or modify it. There is a `FilterDisplayManager` which renders the filters based on the page view. Filters and dividers are passed to it using React props children. On desktop, it displays the filters, and on mobile it collapses them. If the `ViewFiltersButton` is pressed, the filters will expand to fill the whole page.

The mocks show that the filter dividers in the mobile expanded filter view goes to the edge of the screen. In order to accomplish this, I modify the styling of any `Facets`, `StaticFilters`, or `Divider` components inside the `FilterDisplayManager` so that the divider takes up the full width. Unfortunately this means that those components inside the `CollapsibleFiltersDrawer` are forced to use the 'assign' CompositionMethod. It is worth considering alternative approaches in the future on how we can override styling out-of-the-box and also allow the user to provide any kind of css `CompositionMethod`. Also, the Applied Filters need different styling when they are used in the universal results component. This is another use case where the ability to apply custom styling within the search starter and cleanly merge it with the user defined styling would be helpful. For now, I opted to copy the styling into the `SectionHeader` and make it customizable that way. This does introduce a little but of duplication to the styling, however I think that is preferable to modifying to the component on a vertical and having unintended conflicts on the universal page.

J=SLAP-1774
TEST=manual

Test on both mobile and and desktop and various screen sizes. See that on desktop the filters are visible. On mobile, the filters hide. I can press the button to see the filters, and scroll if needed. I can click the close button to go back to the mobile view. Resizing the window from the FiltersVisibleMobile to Desktop will properly update to the Desktop view.


src/components/CollapsibleFilterDrawer.tsx
```diff
@@ -0,0 +1,71 @@
...
+  pageView,
+  setPageView,
+  customCssClasses,
+  cssCompositionMethod
+}: React.PropsWithChildren<Props>) {
+  const cssClasses = useComposedCssClasses(builtInCssClasses, customCssClasses, cssCompositionMethod);
+  if (pageView === PageView.Desktop) {
+    return <div className={cssClasses.container___desktop}>{children}</div>;
+  } else if (pageView === PageView.MobileFiltersExpanded) {
+    const childrenWithWithUpdatedDividers = recursivelyMapChildren(children, child => {
```
nmanu1, line 42: WithWith

src/hooks/usePageView.tsx
```diff
@@ -0,0 +1,44 @@
...
+    function resizeListener() {
+      if (!isMobile()) {
+        setPageView(PageView.Desktop);
+      } else if (pageView === PageView.Desktop) {
+        // Collapse filters when transitioning from Desktop
+        setPageView(PageView.MobileFiltersCollapsed);
+      }
+    }
+
+    window.addEventListener('resize', resizeListener);
```
oshi97, line 37: is it possible to use resize observer here? not sure how hard it would be

src/components/CollapsibleFilterDrawer.tsx
```diff
@@ -0,0 +1,71 @@
...
+  customCssClasses?: CollapsibleFilterDrawerCssClasses,
+  cssCompositionMethod?: CompositionMethod,
+  pageView: PageView,
+  setPageView: (pageView: PageView) => void
+};
+
+/**
+ * Responsible for managing the CollapsibleFilterDrawer based on the current PageView
+ */
+export default function CollapsibleFilterDrawer({
```
oshi97, line 31: what is a "drawer"?

src/hooks/usePageView.tsx
cea2aj: sure, that seems like more of a proper way of doing it

src/components/CollapsibleFilterDrawer.tsx
oshi97: sure

src/hooks/usePageView.tsx
cea2aj: After further investigation it looks like resize observer is used for elements whereas the widow resize event is more commonly used for changes to the whole window, so we're sticking with the window resize event

src/components/CollapsibleFilterContainer.tsx
```diff
@@ -0,0 +1,71 @@
...
+  } else if (pageView === PageView.MobileFiltersExpanded) {
+    const childrenWithUpdatedDividers = recursivelyMapChildren(children, child => {
+      if (child.type !== Facets && child.type !== StaticFilters && child.type !== Divider ) {
+        return child;
+      }
+      return customizeComponentDivider(child, cssClasses.divider);
+    });
+    return (
+      <div className={cssClasses.container___mobileFiltersExpanded}>
+        <CollapseFiltersButton onClick={() => setPageView(PageView.MobileFiltersCollapsed)} />
```
yen-tt, line 50: Can we directly add the button here instead of creating a new React component for it since it doesn't seem to be use anywhere else? Or is there a way to make a generic CollapsibleFiltersButton that can work for collapse and expand behavior with whatever onClick function and css pass in?

src/components/CollapsibleFilterContainer.tsx
```diff
@@ -0,0 +1,71 @@
+import { cloneElement, ReactElement } from "react";
+import { CompositionMethod, useComposedCssClasses } from "../hooks/useComposedCssClasses";
+import { PageView } from "../hooks/usePageView";
+import CollapseFiltersButton from "./CollapseFiltersButton";
+import Facets from "./Facets";
+import StaticFilters, { Divider } from "./StaticFilters";
+import recursivelyMapChildren from "./utils/recursivelyMapChildren";
+
+interface CollapsibleFilterContainerCssClasses {
```
yen-tt, line 9: should this extends 1CollapseFiltersButtonCssClasses`?

tmeyer2115:
> 

What's meant by "Applied Filters need different styling when they are used in the universal results component"? I'm not aware of us displaying Applied Filters on the Universal page. In the Figma mocks, Michael has Filters and Applied Filters under the 'All' tab, but I think that's just a mistake. He didn't update the tab to be "Vertical".

src/components/CollapsibleFilterContainer.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+interface Props {
+  customCssClasses?: CollapsibleFilterContainerCssClasses,
+  cssCompositionMethod?: CompositionMethod,
+  pageView: PageView,
+  setPageView: (pageView: PageView) => void
+};
+
+/**
+ * Responsible for managing the CollapsibleFilterContainer based on the current PageView
```
tmeyer2115, line 29: Can we update the JSDocs here? This reads a bit like a tautology: "The CollapsibleFilterContainer is responsible for managing the CollapsibleFilterContainer"

src/components/CollapsibleFilterContainer.tsx
tmeyer2115: I think moving the `CollapseFiltersButton` into here makes sense. It won't be used elsewhere. If we were to add a `renderToggle` method that creates the button in this function, we could get rid of the `onClick` prop. The `onClick` prop feels a bit strange to me anyway. What happens when you click the button isn't really configurable, it should always collapse or expand.

src/pages/EventsPage.tsx
```diff
@@ -28,40 +31,46 @@ const staticFiltersConfig = [{
...
-      </div>
+      </CollapsibleFilterContainer>
+      { (pageView === PageView.Desktop || pageView === PageView.MobileFiltersCollapsed) &&
+        <div className='flex-grow'>
+          <DirectAnswer />
+          <SpellCheck />
+          <div className='flex'>
+            <ResultsCount />
+            {pageView === PageView.MobileFiltersCollapsed && 
+              <ExpandFiltersButton onClick={() => setPageView(PageView.MobileFiltersExpanded)}/>}
```
tmeyer2115, line 54: Is there any way we could include the expand functionality in the `CollapsibleFilterContainer`? It would mean we don't need to duplicate this code in every vertical page. I think it's also better encapsulation. One Component handles the expansion and collapse of the filters. Now, the logic for this is distributed across multiple areas of the code.

cea2aj:
> > 
> 
> What's meant by "Applied Filters need different styling when they are used in the universal results component"? I'm not aware of us displaying Applied Filters on the Universal page. In the Figma mocks, Michael has Filters and Applied Filters under the 'All' tab, but I think that's just a mistake. He didn't update the tab to be "Vertical".

When the filters are on the universal page next to the section header name, the spacing needs to be different. Applied filters on universal is functionality that we added to the sample app a while ago to match Theme 1.0. When we styled applied filters on vertical, I adjusted the padding so that it worked on universal as well. But you're right, I don't see the mocks for it on universal. I can double check with Michael if we are supposed to display them

src/pages/EventsPage.tsx
cea2aj: Yes, it's possible, but there are a few reasons that lead me to this approach. The CollapsibleFilterContainer is in a different spot on the DOM that the expand filters button. When the  `ExpandFiltersButton` is it's own component, it's free to be placed anywhere on the page. I think it would be odd to have the `ExpandFiltersButton` appear on the page next to `ResultsCount` without the react code indicating that. However if we did want to go with that approach, the entire page would need to be managed by a component with both the `CollapsibleFiltersContainer` and the rest of the page inside of it. Perhaps we could have a `ResultsContainer` to contain the rest of the results page?

Alternatively we could pass the `pageView` to the `ExpandFiltersButton` so that it only renders itself on mobile with filters collapsed, which would let us get rid of that pageView conditional above it.

src/components/CollapsibleFilterContainer.tsx
cea2aj: after moving the button this interface is no longer needed

src/components/CollapsibleFilterContainer.tsx
cea2aj: I'll move the button into the CollapsibleFilterContainer. I actually think it's simplest to get rid of the function to render the button and instead render it right inside the component

src/components/CollapsibleFilterContainer.tsx
```diff
@@ -0,0 +1,78 @@
...
+ */
+export default function CollapsibleFilterContainer({
```
cea2aj, line 33: Per our discussion, I have been thinking of different names for this component. What are your thoughts on `FilterDisplayManager` or `FilterViewManager`? Or maybe even `FilterPresenter`

src/hooks/usePageView.tsx
```diff
@@ -0,0 +1,44 @@
...
+ * @remarks
+ * The page view responds to window resizes and can also be manually set
+ * to change the type of mobile view
+ */
+export function usePageView (): [PageView,  Dispatch<SetStateAction<PageView>>] {
+  const [pageView, setPageView] = useState<PageView>(
+    isMobile() ? PageView.FiltersHiddenMobile : PageView.Desktop);
+
+  function isMobile () {
+    return window.innerWidth <= 768;
```
yen-tt, line 24: I'm wondering if it make sense to create a pageViewContext instead of a hook, and wrap it as a provider in the App component or just call useContext in whatever component/page that requires it without passing the pageView/setPageView field around?

cea2aj: I think that's worth considering since it would let us pass around less data/functions and make it easy for any component to access it. I'll leave it up to @tmeyer2115 if we want to try attempt that now or in a future item

cea2aj: Some more info from the React docs: "Context is primarily used when some data needs to be accessible by many components at different nesting levels. Apply it sparingly because it makes component reuse more difficult."

cea2aj: Updated to use context!

tmeyer2115 (dismissed):

tmeyer2115 (approved):

PR #33, opened 2022-01-11
pagination component and styling

yen-tt:
Add pagination when display verticalResults from vertical page

J=SLAP-1778
TEST=manual

added more event entities to the experience, set limit to 1 in usePageSetupEffect, see that the pagination component display the labels and ellipses as expected

### DIFF
--- a/src/components/VerticalResults.tsx
+++ b/src/components/VerticalResults.tsx
@@ -170,7 +170,8 @@ function Pagination(props: PaginationProps): JSX.Element {
   return (
     <div className={cssClasses.container}>
       <nav className={cssClasses.labelContainer} aria-label="Pagination">
-        <button 
+        <button
+          aria-label='Navigate to the previous results page'
           className={cssClasses.iconContainer}
           onClick={() => executeSearchWithNewOffset(offset - limit)} disabled={pageNumber === 1}
         >
@@ -186,7 +187,8 @@ function Pagination(props: PaginationProps): JSX.Element {
               return <button key={index} className={cssClasses.label} onClick={onSelectNewPage}>{label}</button>
           }
         })}
-        <button 
+        <button
+          aria-label='Navigate to the next results page'
           className={cssClasses.iconContainer}
           onClick={() => executeSearchWithNewOffset(offset + limit)} disabled={pageNumber === maxPageCount}
         >
--- a/src/components/VerticalResults.tsx
+++ b/src/components/VerticalResults.tsx
@@ -146,27 +146,23 @@ function Pagination(props: PaginationProps): JSX.Element {
     newPageNumber && executeSearchWithNewOffset(limit * (newPageNumber - 1));
   }
 
-  const maxPageCount = Math.trunc((numResults - 1) / limit) + 1;
+  const maxPageCount = Math.ceil(numResults / limit);
   const paginationLabels: string[] = [];
-  const pageNumber = (offset/limit) + 1;
+  const pageNumber = (offset / limit) + 1;
   const previousPageNumber = pageNumber - 1;
   const nextPageNumber = pageNumber + 1;
 
-  if (previousPageNumber !== 0) {
-    if (previousPageNumber <= 3) {
-      [...Array(previousPageNumber)].forEach((_, index) => paginationLabels.push(`${index + 1}`));
-    } else {
-      paginationLabels.push('1', '...', `${previousPageNumber}`);
-    }
+  if (previousPageNumber > 3) {
+    paginationLabels.push('1', '...', `${previousPageNumber}`);
+  } else if (previousPageNumber !== 0) {
+    [...Array(previousPageNumber)].forEach((_, index) => paginationLabels.push(`${index + 1}`));
   }
   paginationLabels.push(`${pageNumber}`);
-  if (nextPageNumber <= maxPageCount) {
-    if (maxPageCount - nextPageNumber <= 2 ) {
-      [...Array(maxPageCount - nextPageNumber + 1)]
-        .forEach((_, index) => paginationLabels.push(`${nextPageNumber + index}`));
-    } else {
-      paginationLabels.push(`${nextPageNumber}`, '...', `${maxPageCount}`);
-    }
+  if (maxPageCount - nextPageNumber > 2) {
+    paginationLabels.push(`${nextPageNumber}`, '...', `${maxPageCount}`);
+  } else if (nextPageNumber <= maxPageCount) {
+    [...Array(maxPageCount - nextPageNumber + 1)]
+      .forEach((_, index) => paginationLabels.push(`${nextPageNumber + index}`));
   }
 
   return (
--- a/src/components/VerticalResults.tsx
+++ b/src/components/VerticalResults.tsx
@@ -105,16 +105,18 @@ interface PaginationCssClasses {
   labelContainer?: string,
   label?: string,
   selectedLabel?: string,
-  iconContainer?: string,
+  leftIconContainer?: string,
+  rightIconContainer?: string,
   icon?: string
 }
 
 const builtInPaginationCssClasses: PaginationCssClasses = {
   container: 'flex justify-center mb-4',
-  labelContainer: 'inline-flex rounded-md shadow-sm border border-gray-300 -space-x-px',
-  label: 'relative z-0 inline-flex items-center px-4 py-2 text-sm font-semibold border-l border-r border-gray-300 text-gray-500',
-  selectedLabel: 'relative z-10 inline-flex items-center px-4 py-2 text-sm font-semibold border border-blue-600 text-blue-600',
-  iconContainer: 'inline-flex items-center px-3.5 py-2',
+  labelContainer: 'inline-flex shadow-sm -space-x-px',
+  label: 'z-0 inline-flex items-center px-4 py-2 text-sm font-semibold border border-gray-300 text-gray-500',
+  selectedLabel: 'z-10 inline-flex items-center px-4 py-2 text-sm font-semibold border border-blue-600 text-blue-600 bg-blue-50',
+  leftIconContainer: 'inline-flex items-center px-3.5 py-2 border border-gray-300 rounded-l-md',
+  rightIconContainer: 'inline-flex items-center px-3.5 py-2 border border-gray-300 rounded-r-md',
   icon: 'w-3 text-gray-500'
 }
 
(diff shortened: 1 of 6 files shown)


src/components/VerticalResults.tsx
```diff
@@ -61,19 +62,139 @@ interface VerticalResultsProps {
...
+  label?: string,
+  selectedLabel?: string,
+  iconContainer?: string,
+  icon?: string
+}
+
+const builtInPaginationCssClasses: PaginationCssClasses = {
+  container: 'flex justify-center mb-4',
+  labelContainer: 'inline-flex rounded-md shadow-sm border border-gray-300 -space-x-px',
+  label: 'relative z-0 inline-flex items-center px-4 py-2 text-sm font-semibold border-l border-r border-gray-300 text-gray-500',
```
cea2aj, line 115: What's the reason for using relative?

src/components/VerticalResults.tsx
```diff
@@ -61,19 +62,139 @@ interface VerticalResultsProps {
...
+interface PaginationCssClasses {
+  container?: string,
+  labelContainer?: string,
+  label?: string,
+  selectedLabel?: string,
+  iconContainer?: string,
+  icon?: string
+}
+
+const builtInPaginationCssClasses: PaginationCssClasses = {
```
cea2aj, line 112: Can we update the background of the selected square to be blue to match the mocks?

cea2aj, line 112: Also, small nit, but is is possible to align the blue box with the rest of the pagination border so that it's not inside the other border?

yen-tt: ah right, of course

src/components/VerticalResults.tsx
yen-tt: attempted another approach with it before switching using negative margin, I will get rid of it!

src/components/VerticalResults.tsx
yen-tt: sure thing

src/components/VerticalResults.tsx
```diff
@@ -61,19 +62,141 @@ interface VerticalResultsProps {
...
+    : verticalResultsCount
+
+  const offset = useAnswersState(state => state.vertical.offset) || 0;
+  const limit = useAnswersState(state => state.vertical.limit) || 5;
+
+  return (
+    <>
+      <VerticalResultsDisplay results={results} isLoading={isLoading} {...otherProps}/>
+      {allowPagination 
+        && <Pagination
```
oshi97, line 91: Should pagination be included in the vertical results component?

yen-tt: From tom's description in the item, we would want only VerticalResults component to support Pagination. I think it makes sense this way, since VerticalResults also have conditions that change what kind of results should be use (e.g. normal vertical results for the search or all results for the vertical)

src/components/VerticalResults.tsx
```diff
@@ -61,19 +62,141 @@ interface VerticalResultsProps {
...
+  const executeSearchWithNewOffset = (newOffset: number) => {
+    answersAction.setOffset(newOffset);
+    answersAction.executeVerticalQuery();
+  }
+  const onSelectNewPage = (evt: React.MouseEvent) => {
+    const newPageNumber = Number(evt.currentTarget.textContent);
+    newPageNumber && executeSearchWithNewOffset(limit * (newPageNumber - 1));
+  }
+
+  const maxPageCount = Math.trunc((numResults - 1) / limit) + 1;
```
oshi97, line 149: what is the minus 1 in `numResults - 1` for?

src/components/VerticalResults.tsx
```diff
@@ -61,19 +62,141 @@ interface VerticalResultsProps {
...
+  const paginationLabels: string[] = [];
+  const pageNumber = (offset/limit) + 1;
+  const previousPageNumber = pageNumber - 1;
+  const nextPageNumber = pageNumber + 1;
+
+  if (previousPageNumber !== 0) {
```
oshi97, line 155: maybe this can be
```
if (ppageNum > 3) {
} else if (ppageNum !== 0) {
}
```

src/components/VerticalResults.tsx
```diff
@@ -61,19 +62,141 @@ interface VerticalResultsProps {
...
+    answersAction.executeVerticalQuery();
+  }
+  const onSelectNewPage = (evt: React.MouseEvent) => {
+    const newPageNumber = Number(evt.currentTarget.textContent);
+    newPageNumber && executeSearchWithNewOffset(limit * (newPageNumber - 1));
+  }
+
+  const maxPageCount = Math.trunc((numResults - 1) / limit) + 1;
+  const paginationLabels: string[] = [];
+  const pageNumber = (offset/limit) + 1;
```
oshi97, line 151: does this also need Math.trunc?

src/components/VerticalResults.tsx
yen-tt: to guard the case when `numResults` is exactly the same as `limit`. For example, with numResults and limit as 5,
maxPageCount should be 1 but it would be 2 without the `-1`

src/components/VerticalResults.tsx
```diff
@@ -61,19 +62,141 @@ interface VerticalResultsProps {
...
   const verticalResults = useAnswersState(state => state.vertical.results) || [];
   const allResultsForVertical = useAnswersState(state => state.vertical?.noResults?.allResultsForVertical.results) || [];
+  const verticalResultsCount = useAnswersState(state => state.vertical.resultsCount) || 0;
+  const allResultsCountForVertical = useAnswersState(state => state.vertical?.noResults?.allResultsForVertical.resultsCount) || 0;
   const isLoading = useAnswersState(state => state.searchStatus.isLoading);
 
   const results = verticalResults.length === 0 && displayAllOnNoResults
     ? allResultsForVertical
     : verticalResults
+  const resultsCount = verticalResults.length === 0 && displayAllOnNoResults
```
tmeyer2115, line 80: Minor nit: We do the same conditional check twice in a row:

```
verticalResults.length === 0 && displayAllOnNoResults
```

Could we do the check just once and assign to a variable?

src/components/VerticalResults.tsx
```diff
@@ -61,19 +62,141 @@ interface VerticalResultsProps {
...
+
+  const maxPageCount = Math.trunc((numResults - 1) / limit) + 1;
+  const paginationLabels: string[] = [];
+  const pageNumber = (offset/limit) + 1;
+  const previousPageNumber = pageNumber - 1;
+  const nextPageNumber = pageNumber + 1;
+
+  if (previousPageNumber !== 0) {
+    if (previousPageNumber <= 3) {
+      [...Array(previousPageNumber)].forEach((_, index) => paginationLabels.push(`${index + 1}`));
```
oshi97, line 157: is the spread necessary here? we could alternatively do a normal for loop though it's less fancy

src/components/VerticalResults.tsx
yen-tt: I believe it's not needed. We want pageNumber to start at 1 onward, and with offset always be 0 or multiple of `limit`, this should be ok.

oshi97: I thought we might end up with a decimal on `paginationLabels.push('1', '...', `${previousPageNumber}`);`, but sounds like it's fine?

src/components/VerticalResults.tsx
yen-tt: oh yea, will update!

src/components/VerticalResults.tsx
oshi97: Would `Math.ceil(numResults / limit)` work?

src/components/VerticalResults.tsx
yen-tt: pageNumber will always be a whole number since offset is a multiple of limit. and previousPageNumber is just subtract one of pageNumber so it should be good. Unless I'm not understanding a case you are seeing?

src/components/VerticalResults.tsx
tmeyer2115, line 155: Could we factor out the creation of `paginationLabels` into a separate method? That will make it more unit-testable (when we extract things into a separate Component lib). The logic is somewhat complex, so we will definitely want tests eventually.

src/components/VerticalResults.tsx
yen-tt: oh I think so

src/components/VerticalResults.tsx
oshi97: Oh I didn't realize offset would always be a multiple of the limit. I guess there could be some future case where a different component changes the offset though? It might be simpler to just have this be `Math.ceil(offset/limit)`

src/components/VerticalResults.tsx
```diff
@@ -61,19 +62,141 @@ interface VerticalResultsProps {
...
   const isLoading = useAnswersState(state => state.searchStatus.isLoading);
 
   const results = verticalResults.length === 0 && displayAllOnNoResults
     ? allResultsForVertical
     : verticalResults
+  const resultsCount = verticalResults.length === 0 && displayAllOnNoResults
+    ? allResultsCountForVertical
+    : verticalResultsCount
+
+  const offset = useAnswersState(state => state.vertical.offset) || 0;
```
oshi97, line 84: since these values are only used as props for Pagination, it might make sense to put them in the Pagination component instead of pulling them up. We also don't need these if allowPagination is false

src/components/VerticalResults.tsx
yen-tt: `Math.ceil(offset/limit)` doesn't work since we want the pageNumber to be one ahead of `offset/limit`. So with offset of 4 and limit of 2, we want to be on page 3 instead of 2. From the item, it looks like we only want pagination on vertical results so I think it's ok to keep this as is for now. what do you think?

src/components/VerticalResults.tsx
yen-tt: Array(previousPageNumber) give `[empty × 2]` that's not iterable unless the spread operator is used to fill the values with `undefined`. This approach seems common for people to iterate x times without mutable variables like `i++` (at least from stackoverflow haha). But I can switch if that's prefer

src/components/VerticalResults.tsx
yen-tt: sure thing

tmeyer2115 (approved):

oshi97 (approved):

PR #34, opened 2022-01-11
Nav focus styling tweaks

cea2aj:
Update nav focus so that the focus doesn't collide with the nav bar

In order to accomplish this, I created an extra div so that the nav underline and the button are separate and able to match the mocks.

I also noticed a bug where the menu button would not open if you clicked on the kebab svg, so I fixed that by adding `pointer-events-none`. This makes sure that the parent menu button captures the click and not the SVG

J=SLAP-1776
TEST=manual

Manually tab through the nav and see that the focus box don't collide with the nav bar


src/components/Navigation.tsx
```diff
@@ -91,17 +96,18 @@ export default function Navigation({ links, customCssClasses, cssCompositionMeth
   const { search } = useLocation();
   const visibleLinks = links.slice(0, links.length - numOverflowLinks);
   const overflowLinks = links.slice(-numOverflowLinks);
-  const menuContainsActiveLink = overflowLinks.some(({ to }) =>
-    to === currentVertical || (to === '/' && currentVertical === undefined)
-  );
+  const isActiveLink = ({ to }: LinkData) => to === currentVertical || (to === '/' && currentVertical === undefined)
+  const activeVisibleLinkIndex = visibleLinks.findIndex(isActiveLink);
+  const activeMenuLinkIndex = overflowLinks.findIndex(isActiveLink);
+  const menuContainsActiveLink = overflowLinks.some(isActiveLink);
```
nmanu1, line 102: I think this can be simplified to checking if `activeMenuLinkIndex` is at least 0 or not

src/components/Navigation.tsx
```diff
@@ -127,19 +134,24 @@ export default function Navigation({ links, customCssClasses, cssCompositionMeth
 
 function renderLink(
   linkData: LinkData,
+  isActiveLink: boolean,
   queryParams: string,
-  cssClasses: { navLink?: string, navLink___active?: string }) 
+  cssClasses: { navLinkContainer?: string, navLinkContainer___active?: string, navLink?: string, navLink___active?: string })
```
nmanu1, line 139: I think `navLink___active` can be removed here as well

src/components/Navigation.tsx
```diff
@@ -127,19 +134,24 @@ export default function Navigation({ links, customCssClasses, cssCompositionMeth
...
 function renderLink(
   linkData: LinkData,
+  isActiveLink: boolean,
   queryParams: string,
-  cssClasses: { navLink?: string, navLink___active?: string }) 
+  cssClasses: { navLinkContainer?: string, navLinkContainer___active?: string, navLink?: string, navLink___active?: string }) 
 {
   const { to, label } = linkData;
+  const navLinkContainerClasses = classNames(cssClasses.navLinkContainer, {
+    [cssClasses.navLinkContainer___active ?? '']: isActiveLink
```
nmanu1, line 143: Did we want to switch to using this instead?
```
const navLinkContainerClasses = cssClasses.navLinkContainer___active
  ? classNames(cssClasses.navLinkContainer, {
    [cssClasses.navLinkContainer___active]: isActiveLink
  })
  : cssClasses.navLinkContainer;
```
It avoids the `?? ''` that we were only using above for conciseness and readability because there were multiple classes being checked.

src/components/Navigation.tsx
```diff
@@ -17,20 +19,23 @@ interface NavigationCssClasses {
...
 const builtInCssClasses: NavigationCssClasses = {
   nav: 'border-b border-gray-200 text-gray-600 flex space-x-6 font-medium mb-6',
-  navLink: 'whitespace-nowrap py-3 px-1 mt-1 font-medium text-md border-b-2 border-opacity-0 hover:border-gray-300',
-  navLink___active: 'text-blue-600 border-blue-600 border-b-2 border-opacity-100 hover:border-blue-600',
+  navLinkContainer: 'whitespace-nowrap py-3 mt-1 font-medium text-md border-b-2 border-opacity-0 hover:border-gray-300',
+  navLink: 'py-3 px-2',
+  navLinkContainer___active: 'text-blue-600 border-blue-600 border-opacity-100 hover:border-blue-600',
   menuButtonContainer: 'relative flex flex-grow justify-end mr-4',
-  menuButton: 'flex items-center text-gray-600 font-medium text-md my-1 p-3 border-opacity-0 rounded-md hover:bg-gray-200',
+  menuButton: 'flex items-center text-gray-600 font-medium text-md mt-0.5 p-3 border-opacity-0 rounded-md hover:bg-gray-200',
```
nmanu1, line 32: IIRC changing this will make the background of the menu button (on hover and when open) collide with the nav bar

src/components/Navigation.tsx
```diff
@@ -91,17 +96,18 @@ export default function Navigation({ links, customCssClasses, cssCompositionMeth
   const { search } = useLocation();
```
yen-tt, line 96: what is 'search' here? doesn't seem to be something we need to construct the nav link since the `to` field should already contain the verticalKey. Can we get rid of it?

src/components/Navigation.tsx
```diff
@@ -127,19 +134,24 @@ export default function Navigation({ links, customCssClasses, cssCompositionMeth
...
-      to={`${to}${queryParams}`}
-      exact={true}
-    >
-      {label}
-    </NavLink>
+    <div className={navLinkContainerClasses}>
+      <NavLink
+        key={to}
+        className={cssClasses.navLink}
+        to={`${to}${queryParams}`}
```
yen-tt, line 150: can we get rid of queryParams?

src/components/Navigation.tsx
```diff
@@ -9,6 +9,8 @@ interface NavigationCssClasses {
   nav?: string,
   linksWrapper?: string,
   menuWrapper?: string,
+  navLinkContainer?: string,
+  navLinkContainer___active?: string,
   navLink?: string,
   navLink___active?: string,
```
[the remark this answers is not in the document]
cea2aj: yep, removing!

src/components/Navigation.tsx
cea2aj: ah, your right. setting h-12 on the menuButton seems to fix it

src/components/Navigation.tsx
cea2aj: In my testing, it seems to be empty. I think it's there so that we get the query params when we navigate to another page. However we don't store query params in the URL yet so it's not doing anything right now. I would actually expect `useParams` to be used instead of `useLocation` since we just want the query params. Since I think that's incorrect, I can remove it for now and we can revisit with the query param work

src/components/Navigation.tsx
cea2aj: yep, that's faster too

src/components/Navigation.tsx
cea2aj: removing

src/components/Navigation.tsx
cea2aj: sure, I can update

src/components/Navigation.tsx
cea2aj: sure, removing for now

src/components/Navigation.tsx
nmanu1: the background looks good, but it seems like the 'More' text is slightly higher than the other nav text now

yen-tt (dismissed):

src/components/Navigation.tsx
cea2aj: got it, updating to `mt-1` fixes it

nmanu1 (approved):

PR #40, opened 2022-01-13
Recent searches text matching

nmanu1:
Filter recent searches by whether they are Levenshtein matches for the input query text. Also, updated so that entity previews are displayed when focusing on a recent search.

J=SLAP-1802
TEST=manual

Check that recent searches are properly filtered when typing and entity previews are displayed on focus.

### DIFF
--- a/src/components/VisualAutocomplete/VisualSearchBar.tsx
+++ b/src/components/VisualAutocomplete/VisualSearchBar.tsx
@@ -1,5 +1,5 @@
-import { useAnswersActions, useAnswersState, VerticalResults, AutocompleteResult } from '@yext/answers-headless-react';
-import { PropsWithChildren, useEffect } from 'react';
+import { useAnswersActions, useAnswersState, useAnswersUtilities, VerticalResults, AutocompleteResult } from '@yext/answers-headless-react';
+import { PropsWithChildren, useEffect, useState } from 'react';
 import InputDropdown from '../InputDropdown';
 import '../../sass/Autocomplete.scss';
 import DropdownSection, { Option } from '../DropdownSection';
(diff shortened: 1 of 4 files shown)


src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -90,8 +91,12 @@ export default function VisualSearchBar({
       clearRecentSearches();
     }
   }, [clearRecentSearches, hideRecentSearches])
-  const haveRecentSearches = !hideRecentSearches && recentSearches?.length !== 0;
-  
+
+  const [filteredRecentSearches, setFilteredRecentSearches] = useState(recentSearches);
+  useEffect(() =>
```
yen-tt, line 96: why is this useEffect call needed? I think we only need to call `setFilteredRecentSearches` when onInputFocus (user click or type on searchbar) event is trigger.

nmanu1: I had added this to handle recent searches changing when the dropdown is open (like if recent searches were hidden/cleared). but this seems unlikely now that I think about it more, so I'll remove it

yen-tt (approved):

src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -211,6 +214,9 @@ export default function VisualSearchBar({
         }}
         onInputFocus={value => {
           updateEntityPreviews(value);
+          setFilteredRecentSearches(recentSearches?.filter(search =>
```
cea2aj, line 217: what's the reason for calling this here instead of `onInputChange`?

nmanu1: calling it `onInputFocus` includes the events (i.e. typing) that trigger `onInputChange` as well as triggering when clicking on the input bar. for example, this is needed to filter the recent searches when initially clicking on the input bar (with or without text already present)

src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -211,6 +214,9 @@ export default function VisualSearchBar({
         }}
         onInputFocus={value => {
           updateEntityPreviews(value);
+          setFilteredRecentSearches(recentSearches?.filter(search => 
+            answersUtilities.isCloseMatch(search.query, value)
```
cea2aj, line 218: If someone clicks on the search bar but hasn't typed anything, do we want to show a certain number of recent searches without running the `isCloseMatch` function? Has UX given us any guidance there?

nmanu1: By default, we've been showing up to 5 recent searches when there is no input text, which would also be the case here (all matches would be true for empty search string). I'm not sure if we chose that number or if we got that from UX.

cea2aj: alright, sounds good

cea2aj (approved):

PR #42, opened 2022-01-13
Applied filters update once results load

nmanu1:
Only update applied filters once the results have finished loading. Also, change opacity of applied filters when loading.

J=SLAP-1796
TEST=manual

Check that applied filters have the correct behavior on Locations and Events pages.

### DIFF
--- a/src/components/AppliedFilters.tsx
+++ b/src/components/AppliedFilters.tsx
@@ -31,8 +31,7 @@ interface AppliedFiltersDisplayProps {
   labelText?: string,
   delimiter?: string,
   displayableFilters: DisplayableFilter[],
-  customCssClasses?: AppliedFiltersCssClasses,
-  cssCompositionMethod?: CompositionMethod
+  cssClasses?: AppliedFiltersCssClasses
 }
 
 export interface AppliedFiltersProps {
@@ -57,20 +56,22 @@ export default function AppliedFilters (
     filterState.current = state.vertical.results ? state.filters : {};
   }
 
-  const { hiddenFields = [], staticFiltersGroupLabels = {}, ...otherProps } = props;
+  const { hiddenFields = [], staticFiltersGroupLabels = {}, customCssClasses = {}, cssCompositionMethod, ...otherProps } = props;
   const groupedFilters: Array<GroupedFilters> = getGroupedAppliedFilters(filterState.current, nlpFilters, hiddenFields, staticFiltersGroupLabels);
   const appliedFilters = groupedFilters.flatMap(groupedFilters => groupedFilters.filters);
-  return <AppliedFiltersDisplay displayableFilters={appliedFilters} {...otherProps}/>
+
+  const cssClasses = useComposedCssClasses(builtInCssClasses, customCssClasses, cssCompositionMethod);
+  cssClasses.appliedFiltersContainer = classNames(cssClasses.appliedFiltersContainer, {
+    [cssClasses.appliedFiltersContainer___loading ?? '']: state.searchStatus.isLoading
+  });
+  return <AppliedFiltersDisplay displayableFilters={appliedFilters} cssClasses={cssClasses} {...otherProps}/>
 };
 
 export function AppliedFiltersDisplay ({
   labelText,
   displayableFilters,
-  customCssClasses = {},
-  cssCompositionMethod
+  cssClasses = {}
 }: AppliedFiltersDisplayProps): JSX.Element {
-  const cssClasses = useComposedCssClasses(builtInCssClasses, customCssClasses, cssCompositionMethod);
-  const isLoading = useAnswersState(state => state.searchStatus.isLoading);
 
   function NlpFilter({ filter }: { filter: DisplayableFilter }): JSX.Element {
     return (
@@ -108,16 +109,10 @@ export function AppliedFiltersDisplay ({
     );
   }
 
-  const containerClassNames = cssClasses.appliedFiltersContainer___loading
-    ? classNames(cssClasses.appliedFiltersContainer, {
-      [ cssClasses.appliedFiltersContainer___loading ]: isLoading
-    })
-    : cssClasses.appliedFiltersContainer;
-
   return (
     <>
       { displayableFilters.length > 0 &&
-        <div className={containerClassNames} aria-label={labelText}>
+        <div className={cssClasses.appliedFiltersContainer} aria-label={labelText}>
           {displayableFilters.map((filter: DisplayableFilter) => {
             const key = `${filter.filterType}-${filter.label}`;
             if (filter.filterType === 'NLP_FILTER') {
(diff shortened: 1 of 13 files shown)


tests/appliedFilters.test.tsx
```diff
@@ -95,7 +96,7 @@ describe('AppliedFilters component works as expected', () => {
 
     filerRemoveButton.click();
     facetLabels = container.getElementsByClassName(cssClasses.filterLabel);
-    expect(facetLabels.length).toBe(0);
+    setTimeout(() => expect(facetLabels.length).toBe(0), 1000);
```
cea2aj, line 99: Is there another approach we can use that will let us avoid the `setTimeout`? Using too many `setTimeouts` can increase the total runtime of the tests if they wait longer than necessary

nmanu1: Since the click function for the remove button is synchronous, calling and waiting for `executeVerticalQuery` after clicking the remove button would ensure that `facetLabels` are being checked at the correct time without needing the `setTimeout`. This would just add the extra time it takes to make that additional query

src/components/AppliedFilters.tsx
```diff
@@ -99,10 +108,16 @@ export function AppliedFiltersDisplay ({
     );
   }
 
+  const containerClassNames = cssClasses.appliedFiltersContainer___loading
```
tmeyer2115, line 111: Can we simplify the logic here? It's a bit confusing when we have conditional logic outside the conditional logic within classNames. Can we use this pattern to clean things up:

```
[ cssClasses.appledFiltersContainer__loading || '' ]: isLoading
```

src/components/AppliedFilters.tsx
```diff
@@ -62,6 +70,7 @@ export function AppliedFiltersDisplay ({
   cssCompositionMethod
 }: AppliedFiltersDisplayProps): JSX.Element {
   const cssClasses = useComposedCssClasses(builtInCssClasses, customCssClasses, cssCompositionMethod);
+  const isLoading = useAnswersState(state => state.searchStatus.isLoading);
```
tmeyer2115, line 73: The Display Components we write are supposed to be pretty dumb. Because of that, I'd like to avoid having `useAnswersState` calls in them if we can. It makes the Components simpler, easier to reuse, and easier to test. They render DOM purely based on the props given them. 

In this case, I'd suggest we either pass `isLoading` as a prop to the Component or we pass in a `cssClasses` prop. This latter prop would make the Component even simpler. We'd handle the `useComposedCssClasses` upstream. Additionally, we could modify `cssClasses.appliedFiltersContainer` upstream to include the `isLoading` conditional. Then, this Component wouldn't need to have any knowledge of the loading state at all.

nmanu1: It might also just be simpler to set loading opacity at a higher level rather than by component. If other components on each page (spellcheck, result count, etc.) also should be at half opacity when results are loading, this could be set on each page to encompass all of those components.

cea2aj: I like the idea of applying the opacity at a higher level. Potentially we could have a component which listens to `searchStatus.isLoading` and takes in some children components and renders them all within a div with opacity at 50%. This would be out of scope of this item though. What do you think @tmeyer2115 , should we make a new item for that?

src/components/AppliedFilters.tsx
nmanu1: Updated!

src/components/AppliedFilters.tsx
nmanu1: I've added the opacity for `AppliedFilters` for now, but applying it at a higher level will be done in a separate item after more discussion on which components should be at half opacity while results are loading. For this PR, I've simplified `AppliedFiltersDisplay` so it only takes in a `cssClasses` prop instead of composing the classes itself.

src/components/AppliedFilters.tsx
```diff
@@ -48,20 +51,27 @@ export default function AppliedFilters (
 ): JSX.Element {
   const nlpFilters = useAnswersState(state => state.vertical?.appliedQueryFilters) || [];
   const state = useAnswersState(state => state);
-  const filterState = state.vertical.results ? state.filters : {};
-  const { hiddenFields = [], staticFiltersGroupLabels = {}, ...otherProps } = props;
-  const groupedFilters: Array<GroupedFilters> = getGroupedAppliedFilters(filterState, nlpFilters, hiddenFields, staticFiltersGroupLabels);
+  const filterState = useRef<FiltersState>({});
```
tmeyer2115, line 54: I'm wondering if we can avoid the use of a Ref here. In general, we should use Refs as a matter of last resort. They aren't exactly an anti-pattern, but they're counter to the declarative nature of React. I'm wondering if we can do something like:

```
const canDisplayFilters = useAnswersState(state => !state.searchStatus.isLoading && state.vertical.results);
const filters = useAnswersState(state => state.filters);
const displayedFilters = canDisplayFilters ? filters : {};
```

nmanu1: The reason I used Ref here was to continue showing the last displayed filters until the new results load. With this approach, the previous filters would be removed while the results are loading and then the new ones would be displayed, instead of persisting the old filters with the old results

src/components/AppliedFilters.tsx
```diff
@@ -48,20 +51,27 @@ export default function AppliedFilters (
 ): JSX.Element {
   const nlpFilters = useAnswersState(state => state.vertical?.appliedQueryFilters) || [];
   const state = useAnswersState(state => state);
```
oshi97, line 53: little out of scope but I don't think we need to watch the entire state tree?

nmanu1: yeah that's true. I'll update it

src/components/AppliedFilters.tsx
```diff
@@ -46,22 +49,32 @@ export interface AppliedFiltersProps {
...
-  const { hiddenFields = [], staticFiltersGroupLabels = {}, ...otherProps } = props;
-  const groupedFilters: Array<GroupedFilters> = getGroupedAppliedFilters(filterState, nlpFilters, hiddenFields, staticFiltersGroupLabels);
+  const nlpFilters = useAnswersState(state => state.vertical.appliedQueryFilters) || [];
+  const isLoading = useAnswersState(state => state.searchStatus.isLoading);
+  const verticalResults = useAnswersState(state => state.vertical.results);
+  const filters = useAnswersState(state => state.filters);
+
+  const filterState = useRef<FiltersState>({});
+  if (!isLoading) {
+    filterState.current = verticalResults ? filters : {};
```
oshi97, line 59: why do we set filtersState.current to verticalResults?

nmanu1: this is setting filterState.current to the current filters, unless there are no vertical results, in which case the displayed applied filters are empty

oshi97: does `filtersState.current = filters` not work?

nmanu1: no, this would display the applied filters when there are no vertical results returned. I assume the expected behavior is to have applied filters not displayed for no results since that's how it was before this item. is that not the case?

oshi97 (approved):
lgtm!

PR #43, opened 2022-01-14
Include recent searches option in screen reader

yen-tt:
Update screen reader text in search bar to include number of recent search options, and announce the option type when navigating through the option (e.g. a recent search or an autocomplete option)

J=SLAP-1792
TEST=manual

turn on screen reader on Mac, see that the screen reader announced the right number and type of option for recent searches and autocomplete option.

### DIFF
--- a/src/components/VisualAutocomplete/VisualSearchBar.tsx
+++ b/src/components/VisualAutocomplete/VisualSearchBar.tsx
@@ -1,4 +1,4 @@
-import { useAnswersActions, useAnswersState, useAnswersUtilities, VerticalResults, AutocompleteResult } from '@yext/answers-headless-react';
+import { useAnswersActions, useAnswersState, useAnswersUtilities, VerticalResults } from '@yext/answers-headless-react';
 import { PropsWithChildren, useEffect, useState } from 'react';
 import InputDropdown from '../InputDropdown';
 import '../../sass/Autocomplete.scss';
@@ -129,7 +129,12 @@ export default function VisualSearchBar({
           answersActions.setQuery(result.value);
           executeQuery();
         },
-        display: renderAutocompleteResult(result, cssClasses, MagnifyingGlassIcon)
+        display: renderAutocompleteResult(
+          result,
+          cssClasses,
+          MagnifyingGlassIcon,
+          `autocomplete option: ${result.value}`
+        )
       });
 
       if (hideVerticalLinks) {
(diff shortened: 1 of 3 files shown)


src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -241,10 +251,18 @@ export default function VisualSearchBar({
   )
 }
 
-function getScreenReaderText(options: AutocompleteResult[]) {
-  return processTranslation({
-    phrase: `${options.length} autocomplete option found.`,
-    pluralForm: `${options.length} autocomplete options found.`,
-    count: options.length
+function getScreenReaderText(autocompleteOptions: number, recentSearchesOptions: number) {
```
nmanu1, line 254: do we also need to announce the number of entity previews? if we're allowing those to be navigable by arrow key, it might be confusing that you can keep going in the dropdown after surpassing the number of things announced

src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -241,10 +251,18 @@ export default function VisualSearchBar({
...
+      pluralForm: `${recentSearchesOptions} recent search options found.`,
+      count: recentSearchesOptions
+    })
+    : '';
+  const autocompleteText = processTranslation({
+    phrase: `${autocompleteOptions} autocomplete option found.`,
+    pluralForm: `${autocompleteOptions} autocomplete options found.`,
+    count: autocompleteOptions
   });
+  return recentSearchesText + ' ' + autocompleteText;
```
nmanu1, line 267: can we guard this so the space is not added when it's not needed (e.g. no recent searches)?

src/components/VisualAutocomplete/VisualSearchBar.tsx
yen-tt: currently, I don't believe the visual search bar we have in the sample app can navigate through entity previews. So there may be a case we announce number of entity previews but it's not be navigable, unless I suppose user modify the render function somehow to use a DropdownSection. Would it be less confusing if number of entity previews is included regardless if it's navigable or not?

src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -129,7 +129,12 @@ export default function VisualSearchBar({
           answersActions.setQuery(result.value);
           executeQuery();
         },
-        display: renderAutocompleteResult(result, cssClasses, MagnifyingGlassIcon)
+        display: renderAutocompleteResult(
+          result,
+          cssClasses,
+          MagnifyingGlassIcon,
+          `autocomplete option: ${result.value}`
```
cea2aj, line 136: nit: What do you think of using 'autocomplete suggestion:' and 'recent search:' instead?

src/components/VisualAutocomplete/VisualSearchBar.tsx
nmanu1: maybe this is something that should wait until the item that adds arrow key navigation for the entity previews. we'll have a better idea then if it's something we would always announce or not

yen-tt: sure, I will resolve this comment then

cea2aj (approved):

PR #44, opened 2022-01-18
Update query source

nmanu1:
Use the "AUTOCOMPLETE" query source when hitting the universal endpoint from visual autocomplete to get entity previews and when hitting the vertical endpoint from clicking on a vertical link. Also add the query param to the URL so refreshing and browser history is accurate.

J=SLAP-1801
TEST=manual

Check that requests from VA entity previews and from clicking vertical links have `source=AUTOCOMPLETE` but other searches from VA or the regular search bar have the original source. Also check that the URL query param works as expected.

### DIFF
--- a/src/components/VisualAutocomplete/VisualSearchBar.tsx
+++ b/src/components/VisualAutocomplete/VisualSearchBar.tsx
@@ -81,7 +81,6 @@ export default function VisualSearchBar({
   const answersActions = useAnswersActions();
   const answersUtilities = useAnswersUtilities();
   const query = useAnswersState(state => state.query.input) ?? '';
-  const querySource = useAnswersState(state => state.query.querySource);
   const isLoading = useAnswersState(state => state.searchStatus.isLoading) ?? false;
   const [executeQueryWithNearMeHandling, autocompletePromiseRef] = useSearchWithNearMeHandling(answersActions);
   const [autocompleteResponse, executeAutocomplete] = useSynchronizedRequest(async () => {
@@ -156,9 +155,8 @@ export default function VisualSearchBar({
           onSelect: () => {
             autocompletePromiseRef.current = undefined;
             answersActions.setQuery(result.value);
-            answersActions.setQuerySource(QuerySource.Autocomplete);
             browserHistory.push(`/${link.verticalKey}`, {
-              originalQuerySource: querySource
+              querySource: QuerySource.Autocomplete
             });
           },
           display: renderAutocompleteResult({ value: `in ${link.label}` }, { ...cssClasses, option: cssClasses.verticalLink })
--- a/src/components/VisualAutocomplete/VisualSearchBar.tsx
+++ b/src/components/VisualAutocomplete/VisualSearchBar.tsx
@@ -1,4 +1,4 @@
-import { useAnswersActions, useAnswersState, useAnswersUtilities, VerticalResults } from '@yext/answers-headless-react';
+import { useAnswersActions, useAnswersState, useAnswersUtilities, QuerySource, VerticalResults } from '@yext/answers-headless-react';
 import { PropsWithChildren, useEffect, useState } from 'react';
 import InputDropdown from '../InputDropdown';
 import '../../sass/Autocomplete.scss';
@@ -17,6 +17,7 @@ import { ReactComponent as RecentSearchIcon } from '../../icons/history.svg';
 import useRecentSearches from '../../hooks/useRecentSearches';
 import { useHistory } from 'react-router';
 import { ReactComponent as MagnifyingGlassIcon } from '../../icons/magnifying_glass.svg';
+import { BrowserState } from '../../PageRouter';
 
 const SCREENREADER_INSTRUCTIONS = 'When autocomplete results are available, use up and down arrows to review and enter to select.'
 const builtInCssClasses: VisualSearchBarCssClasses = { 
@@ -76,10 +77,11 @@ export default function VisualSearchBar({
 }: PropsWithChildren<Props>) {
   const cssClasses = useComposedCssClasses(builtInCssClasses, customCssClasses, cssCompositionMethod);
 
-  const browserHistory = useHistory();
+  const browserHistory = useHistory<BrowserState>();
   const answersActions = useAnswersActions();
   const answersUtilities = useAnswersUtilities();
   const query = useAnswersState(state => state.query.input) ?? '';
+  const querySource = useAnswersState(state => state.query.querySource);
   const isLoading = useAnswersState(state => state.searchStatus.isLoading) ?? false;
   const [executeQueryWithNearMeHandling, autocompletePromiseRef] = useSearchWithNearMeHandling(answersActions);
   const [autocompleteResponse, executeAutocomplete] = useSynchronizedRequest(async () => {
@@ -154,8 +156,10 @@ export default function VisualSearchBar({
           onSelect: () => {
             autocompletePromiseRef.current = undefined;
             answersActions.setQuery(result.value);
-            executeQuery();
-            browserHistory.push(`/${link.verticalKey}`);
+            answersActions.setQuerySource(QuerySource.Autocomplete);
+            browserHistory.push(`/${link.verticalKey}`, {
+              originalQuerySource: querySource
+            });
           },
           display: renderAutocompleteResult({ value: `in ${link.label}` }, { ...cssClasses, option: cssClasses.verticalLink })
         })
--- a/src/components/VisualAutocomplete/VisualSearchBar.tsx
+++ b/src/components/VisualAutocomplete/VisualSearchBar.tsx
@@ -155,7 +155,7 @@ export default function VisualSearchBar({
           onSelect: () => {
             autocompletePromiseRef.current = undefined;
             answersActions.setQuery(result.value);
-            browserHistory.push(`/${link.verticalKey}`, {
+            browserHistory.push(`/${link.verticalKey}?query=${result.value}`, {
               querySource: QuerySource.Autocomplete
             });
           },
--- a/src/hooks/useEntityPreviews.tsx
+++ b/src/hooks/useEntityPreviews.tsx
@@ -1,5 +1,5 @@
 import { useRef, useState } from "react";
-import { provideAnswersHeadless, useAnswersState, VerticalResults, AnswersHeadless, UniversalLimit, QuerySource } from '@yext/answers-headless-react';
+import { provideAnswersHeadless, VerticalResults, AnswersHeadless, UniversalLimit, QuerySource } from '@yext/answers-headless-react';
 import { answersHeadlessConfig } from '../config/answersHeadlessConfig';
 import useDebouncedFunction from './useDebouncedFunction';
 import useComponentMountStatus from "./useComponentMountStatus";
--- a/src/hooks/useEntityPreviews.tsx
+++ b/src/hooks/useEntityPreviews.tsx
@@ -25,15 +25,14 @@ export function useEntityPreviews(headlessId: string, debounceTime: number):[ En
       ...answersHeadlessConfig,
       headlessId
     });
+    headlessRef.current.setQuerySource(QuerySource.Autocomplete);
   }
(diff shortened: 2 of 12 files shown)


tmeyer2115 (commented):
According to the spec: 

```
When the /universal endpoint is hit from the autocomplete component, pass source=AUTOCOMPLETE. This should also be passed when clicking a Vertical Link, which will navigate to a vertical page. 
```

Will the Vertical Link updates be made in a different PR?

nmanu1:
> According to the spec:
> 
> ```
> When the /universal endpoint is hit from the autocomplete component, pass source=AUTOCOMPLETE. This should also be passed when clicking a Vertical Link, which will navigate to a vertical page. 
> ```
> 
> Will the Vertical Link updates be made in a different PR?

Updated to also change the source for vertical links

src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -154,8 +156,10 @@ export default function VisualSearchBar({
           onSelect: () => {
             autocompletePromiseRef.current = undefined;
             answersActions.setQuery(result.value);
-            executeQuery();
-            browserHistory.push(`/${link.verticalKey}`);
+            answersActions.setQuerySource(QuerySource.Autocomplete);
+            browserHistory.push(`/${link.verticalKey}`, {
+              originalQuerySource: querySource
```
yen-tt, line 161: I think we can just directly send `originalQuerySource: QuerySource.Autocomplete` without needing to listen on querySource or setting it.

src/hooks/usePageSetupEffect.tsx
```diff
@@ -28,7 +31,10 @@ export default function usePageSetupEffect(verticalKey?: string) {
         await updateLocationIfNeeded(answersActions, searchIntents);
       }
       executeSearch(answersActions, !!verticalKey);
+      if (browserLocation.state) {
+        answersActions.setQuerySource(browserLocation.state.originalQuerySource ?? QuerySource.Standard);
```
yen-tt, line 35: can this be `answersActions.setQuerySource(browserLocation.state?.originalQuerySource ?? QuerySource.Standard);`?

src/hooks/usePageSetupEffect.tsx
```diff
@@ -28,7 +31,10 @@ export default function usePageSetupEffect(verticalKey?: string) {
...
+      }
```
yen-tt, line 36: can this be `answersActions.setQuerySource(browserLocation.state?.originalQuerySource ?? QuerySource.Standard);`?

src/components/VisualAutocomplete/VisualSearchBar.tsx
nmanu1: `originalQuerySource` tracks what the source used to be before it was changed to Autocomplete (so either Standard, Overlay, or undefined). that way, it can be set back to this source after finishing setup for the vertical page

src/hooks/usePageSetupEffect.tsx
nmanu1: this would work too, but it would call `setQuerySource` every time a page is set up, even when not coming from a vertical link. having it the way it currently is avoids extra calls for resetting the query source when it hasn't been changed to Autocomplete.

but, it isn't a lot of extra overhead to change it to this and just call it every time a page is set up if that's preferred

src/components/VisualAutocomplete/VisualSearchBar.tsx
yen-tt: I think it would better to keep this logic in one place? So only usePageEffect manage the before/after of querySource, instead of spitting the process between here and usePageEffect. We can directly send QuerySource.Autocomplete through browserHistory. Then in usePageSetupEffect, either execute query normally or, setQuerySource to new query source, execute query, and set it back to old one.

src/hooks/usePageSetupEffect.tsx
yen-tt: actually, out of curiosity, did you check if browserLocation.state would change/reset after first query or when you navigate back and forth in the browser pages?

nmanu1: yes, navigating to other vertical pages resets the browserLocation.state

src/components/VisualAutocomplete/VisualSearchBar.tsx
cea2aj: What's the reason for using browser history here? Could we set the querySource, execute, the query, and then set it back?

nmanu1: Browser history is needed for the vertical links so they navigate to a new vertical page. The changes here pass data along when updating the history to indicate the page change was triggered from VA.

I can consolidate the logic so that both source changes are in `usePageEffect`.

cea2aj: I spoke with Nidhi and the issue is that the `usePageSetupEffect` runs a search any time the vertical changes. This means that if we execute the query in `onSelect` here, we'll fire an extra query. If we want to run the query in `onSelect` without firing the extra query, we'd need to update `usePageSetupEffect` to no longer fire a query when the vertical changes. This would have implications outside the scope of this item because we would need to manually fire a search for regular page load, alternative verticals links, and view more links. If we do want to rework the query firing in `usePageSetupEffect` I'd argue that's out of scope of this item.

A potential benefit of the browser history approach is that the query source info is maintained even if someone refreshes the page. If someone clicks a vertical link, the `AUTOCOMPLETE` query source will be used, and it will keep using it if someone refreshes the page. Rose is out, but I think she might prefer this behavior since it indicates that the API came up with the search term and not the user.

src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -154,8 +155,9 @@ export default function VisualSearchBar({
           onSelect: () => {
             autocompletePromiseRef.current = undefined;
             answersActions.setQuery(result.value);
-            executeQuery();
-            browserHistory.push(`/${link.verticalKey}`);
+            browserHistory.push(`/${link.verticalKey}`, {
```
cea2aj, line 158: Can we also put the query in the URL just like the section header does for the view all links? That way refreshing the page after clicking a link won't result in blank searches with the querySource of 'AUTOCOMPLETE'

src/hooks/useEntityPreviews.tsx
```diff
@@ -28,10 +28,12 @@ export function useEntityPreviews(headlessId: string, debounceTime: number):[ En
   }
   const isMountedRef = useComponentMountStatus();
   const [verticalResultsArray, setVerticalResultsArray] = useState<VerticalResults[]>([]);
+  const querySource = useAnswersState(state => state.query.querySource);
```
oshi97, line 31: for my edification why do we need the main headless's querySource?

src/hooks/useEntityPreviews.tsx
```diff
@@ -28,10 +28,12 @@ export function useEntityPreviews(headlessId: string, debounceTime: number):[ En
...
   const debouncedUniversalSearch = useDebouncedFunction(async () => {
     if (!headlessRef.current) {
       return;
     }
+    headlessRef.current.setQuerySource(QuerySource.Autocomplete);
```
oshi97, line 36: does the query source need to be re-set to Autocomplete every time, or could we set it to Autocomplete once when Headless is initialized?

src/hooks/usePageSetupEffect.tsx
```diff
@@ -21,14 +24,33 @@ export default function usePageSetupEffect(verticalKey?: string) {
...
+        const queryIndex = browserLocation.search.indexOf(queryParamTerm);
+        if (queryIndex >= 0) {
+          const encodedQuery = browserLocation.search.slice(queryIndex + queryParamTerm.length);
+          answersActions.setQuery(decodeURI(encodedQuery));
+        }
+      }
+
+      if (browserLocation.state?.querySource) {
+        const querySource = answersActions.state.query.querySource;
+        answersActions.setQuerySource(browserLocation.state.querySource);
```
oshi97, line 46: do we need to story querySource in the history state? I thought the only time it wouldn't be set to STANDARD is if it comes from the `useEntityPreviews` hook

src/hooks/useEntityPreviews.tsx
nmanu1: the querySource should only be Autocomplete for the request sent from here (and vertical links) in VA. once this request is sent, the source should be switched back to main headless's querySource so further requests don't come from Autocomplete.

for example, if the user types something, the source for hitting the universal endpoint to update entity previews will have Autocomplete as the source, but at any point, if the user presses Enter and triggers the actual search, the source has been switched back to what it originally was for this and any following requests

src/hooks/useEntityPreviews.tsx
nmanu1: good point, I can just set it at the beginning

src/hooks/usePageSetupEffect.tsx
nmanu1: querySource is not stored in the history state except for if a vertical link is clicked. Autocomplete is being stored in that case so `usePageSetupEffect` knows that switching to the vertical page was triggered from VA and it sets the querySource accordingly for that request (and then switches it back for any following requests)

src/hooks/usePageSetupEffect.tsx
```diff
@@ -21,14 +24,33 @@ export default function usePageSetupEffect(verticalKey?: string) {
...
       let searchIntents: SearchIntent[] = [];
       if (!answersActions.state.location.userLocation) {
         searchIntents = await getSearchIntents(answersActions, !!verticalKey) || [];
         await updateLocationIfNeeded(answersActions, searchIntents);
       }
-      executeSearch(answersActions, !!verticalKey);
+
+      if (browserLocation.search) {
+        const queryParamTerm = 'query='
+        const queryIndex = browserLocation.search.indexOf(queryParamTerm);
```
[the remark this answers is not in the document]
nmanu1: updated to use URLSearchParams!

src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -154,8 +155,9 @@ export default function VisualSearchBar({
           onSelect: () => {
             autocompletePromiseRef.current = undefined;
             answersActions.setQuery(result.value);
-            executeQuery();
-            browserHistory.push(`/${link.verticalKey}`);
+            browserHistory.push(`/${link.verticalKey}?query=${result.value}`, {
```
oshi97, line 158: since this is a single page app is it possible to set querySource state here instead of add it to the next browser state?

oshi97: assuming this is possible I'm not sure which is better

nmanu1: it is possible to set the querySource to Autocomplete here. but then we'd still need to know which querySource to switch back to after firing the request in `usePageSetupEffect` (since only that one request is coming from VA and not following ones). so the original querySource would have to be sent in the browser history so it can be set back after the request is fired. I originally used this method, but Yen pointed out that it might be good to have all of the logic in one place in `usePageSetupEffect`, so I updated that in the "consolidate logic" commit

nmanu1: it should be set to whatever the querySource was before it was AUTOCOMPLETE (either STANDARD or OVERLAY)

src/components/VisualAutocomplete/VisualSearchBar.tsx
nmanu1: Added the query param in the url here and throughout the app

cea2aj (approved):
looks good!

src/components/VisualAutocomplete/VisualSearchBar.tsx
oshi97: ah ok got it! I didn't know about OVERLAY

PR #45, opened 2022-01-18
Add Dropdown compound component

oshi97:
This component is based on the dropdown components from [react-bootstrap](https://react-bootstrap.github.io/components/dropdowns/) and [restart/ui](https://react-restart.github.io/ui/Dropdown), which are uncoincidentally maintained by mostly the same people.

The useRootClose and useGlobalListener hooks are taken from directly from restart/ui.
- useRootClose gives modal like behavior to the dropdown, so that the dropdown closes when a click
is done outside of it, and also when the escape key is pressed.
- useGlobalListener adds a global listener to the document.

We could not copy react-bootstrap's approach for maintaining focus on a particular DropdownItem, which under the hood uses restart/ui's base Dropdown component.
This is because restart/ui's base Dropdown maintains focus by using the browser's native focus functionality.
This works great when you don't have an input, but with an input we want the browser's focus to be on that input 100% of the time, and the arrow keys to navigate a sort of "pseudo input"

This was remedied using a `numItems` prop, representing the total number of DropdownItems to be rendered, and also assigning each instance of DropdownItem an `index` prop. An alternative solution would be to decorate props.children like InputDropdown currently does.

J=SLAP-1777
TEST=manual

refactored the SearchBar to use the new Dropdown components (not part of this PR)
can navigate items
can loop through items using the arrow keys (both up and down)
autocomplete works
near me searches work
can run searches
did not notice any styling regressions (aimed to make DOM structure match the current InputDropdown)

did not test visual autocomplete

screenreader reads the annoucement text when the searchbar is clicked
used this [screenreader](https://chrome.google.com/webstore/detail/screen-reader/kgejglhpjiefppelpmljglcjbhoiplfn) (thanks @nmanu1 ) 
screenreader reads the SCREENREADER_INSTRUCTIONS only when clicked on and not when navigating dropdown items

----------

more notes on restart/ui's dropdown:
In restart/ui,
[First](https://github.com/react-restart/ui/blob/adf9d2d4835756b62e842f5567eb4be744291dce/src/Dropdown.tsx#L297), a native `querySelectorAll` is ran for all child elements with a specific magic data- attribute, and the component gets the element in the NodeList with an index of one more (or less, depending on which arrow key is pressed) than the element that currently has focus.
[Then](https://github.com/react-restart/ui/blob/adf9d2d4835756b62e842f5567eb4be744291dce/src/Dropdown.tsx#L298), the element is manually focused.

This does not work with a SearchBar because we want the browser's focus to be maintained on the `input` element 100% of the time. Otherwise, pressing an arrow key would result in leaving the input and not being able to type anymore.
We need to be able to maintain a "pseudo-focus" state like we currently do with InputDropdown.


src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const inputContext: InputContextType = useMemo(() => ({
+    value,
+    setValue,
+    lastTypedOrSubmittedValue,
+    setLastTypedOrSubmittedValue
+  }), [value, lastTypedOrSubmittedValue]);
+
+  const [focusedIndex, setFocusedIndex] = useState(-1);
+  const [focusedValue, setFocusedValue] = useState<string | null>(null);
+  const focusContext: FocusContextType = useMemo(() => ({
```
tmeyer2115, line 47: Here and in the `useMemo` call above, how much of a performance gain are we really getting? Are the `useState` calls really that expensive on a re-render? In short, I'm wondering if the optimization here is really worth the code. Adding all these `useMemo` calls detracts from the readability of this file and make things a bit more confusing.

src/components/Dropdown/DropdownInput.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * An input component for use within a Dropdown.
+ */
+export default function DropdownInput(props: {
+  className?: string,
+  placeholder?: string,
+  onSubmit?: (value?: string) => void,
+  onFocus?: (value?: string) => void,
+  onType?: (value?: string) => void
```
tmeyer2115, line 15: Is `onType` the same thing as `onChange`? If so, I think we should rename it to the latter. I think that's the more common terminology.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,113 @@
...
+  activeClassName?: string,
+  onFocusChange?: (index?: number) => void
+}>) {
+  const {
+    children,
+    numItems,
+    screenReaderUUID,
+    initialValue,
+    onSelect,
+    as: ContainerElementType = 'div',
```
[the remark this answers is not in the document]
oshi97: This is something I saw in the react-bootstrap code, and it lets you specify the kind of element used for the container. Here it defaults to div, but if you wanted the dropdown to be say a `<ul>` to level element, you would do `<Dropdown as='ul'>`. On second thought we don't have any need for this right now, so I'll remove it!
This is the article I originally read about it https://levelup.gitconnected.com/create-more-extensible-react-components-with-the-as-prop-pattern-b79bcbcf4024

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * Dropdown is the parent component for a set of Dropdown-related components.
+ *
+ * It provides multiple shared contexts, which are consumed by its child components,
+ * and also registers some global event listeners.
+ */
+export default function Dropdown(props: PropsWithChildren<{
+  numItems: number,
+  screenReaderUUID?: string,
```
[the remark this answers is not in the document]
oshi97: Originally, I didn't want `Dropdown` to prescribe how the `ScreenReader` component is used. On second thought, though, the main benefit of not prescribing something is greater flexibility, but I can't really think of what kind of flexibility somebody would want. I think I can bake it into `Dropdown` and pull it out later if we need

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const focusContext: FocusContextType = useMemo(() => ({
+    focusedIndex,
+    setFocusedIndex,
+    focusedValue,
+    setFocusedValue
+  }), [focusedIndex, focusedValue]);
+
+  const [_isOpen, _toggle] = useState(false);
+  const isOpen = _isOpen && numItems > 0;
+  const decoratedToggle = useCallback((nextIsOpen: boolean) => {
```
yen-tt: nit: can we rename `nextIsOpen`? and if we are using `_isOpen` with `isOpen`, can we also do `_toggle` and `toggle` instead of `decoratedToggle`?

src/components/Dropdown/Dropdown.tsx
oshi97: It doesn't seem significant, for example with useMemo/Callback hitting the down arrow key to navigate the focus state would result in 3 renders instead of 6. I can remove

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,122 @@
...
+    }
+    if (e.key === 'ArrowDown') {
+      e.preventDefault();
+      const updatedFocusedIndex = focusedIndex + 1;
+      setFocusedValue(null);
+      if (updatedFocusedIndex >= numItems) {
+        setFocusedIndex(-1);
+        setValue(lastTypedOrSubmittedValue);
+      } else {
+        setFocusedIndex((focusedIndex + 1) % numItems);
```
yen-tt, line 87: small thing, can this be `setFocusedIndex(updatedFocusedIndex % numItems);` instead?

src/components/Dropdown/DropdownInput.tsx
oshi97: I was a little concerned with `onChange` not being fired when the inputs value changes due to focusing an autocomplete suggestion. So, hitting the down arrow key to focus a suggestion, which will change the value of the input, won't actually trigger `onChange`. I'm okay with this caveat though

src/components/Dropdown/DropdownInput.tsx
```diff
@@ -0,0 +1,74 @@
...
+    onSubmit,
+    onFocus,
+    onType
+  } = props;
+
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const { decoratedToggle, onSelect, screenReaderUUID } = useContext(DropdownContext) || {};
+  const { value = '', setValue, setLastTypedOrSubmittedValue } = useContext(InputContext) || {};
+  const { focusedIndex = -1, setFocusedIndex } = useContext(FocusContext) || {};
```
yen-tt, line 29: Should these compound component be use outside of the parent? If not, maybe we should make `useDropdownContext`, `useInputContext`, `useFocusContext` so it can guard the usage of compound component (throw error if React.useContext(BlahContext) is undefined)?
Otherwise, can we have default values for these fields or a predefined object pass to the corresponding createContext calls so we don't need to have checks when invoking these fields (blahFunction && blahFunction(...))?

src/components/Dropdown/Dropdown.tsx
oshi97: updated!

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,111 @@
...
+ * It provides multiple shared contexts, which are consumed by its child components,
+ * and also registers some global event listeners.
+ */
+export default function Dropdown(props: PropsWithChildren<{
+  numItems: number,
+  screenReaderUUID?: string,
+  initialValue?: string,
+  onSelect?: (value?: string, index?: number) => void,
+  className?: string,
+  activeClassName?: string,
```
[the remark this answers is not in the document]
oshi97: I think here I prefer two separate props. It's unlikely we're ever going to need additional css classes, since the DOM structure of Dropdown in theory should never change (it should only ever prescribe a single div + the `ScreenReader` which shouldn't ever be visible). Having a prop that's an object makes it a little harder to ensure reference equality for that prop, and the parent component has to wrap it in a useMemo or as a constant outside of the component, and also can't specify the values inline.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,122 @@
...
+    }
+    if (e.key === 'ArrowDown') {
+      e.preventDefault();
+      const updatedFocusedIndex = focusedIndex + 1;
+      setFocusedValue(null);
+      if (updatedFocusedIndex >= numItems) {
+        setFocusedIndex(-1);
+        setValue(lastTypedOrSubmittedValue);
+      } else {
+        setFocusedIndex(updatedFocusedIndex % numItems);
```
cea2aj, line 87: The mod shouldn't affect this, correct?

src/components/Dropdown/DropdownInput.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  const handleKeyDown = (e: KeyboardEvent<HTMLInputElement>) => {
+    if (e.key === 'Enter') {
+      decoratedToggle && decoratedToggle(false);
+      onSubmit && onSubmit(value);
+      setLastTypedOrSubmittedValue && setLastTypedOrSubmittedValue(value);
+      if (focusedIndex >= 0) {
+        onSelect && onSelect(value, focusedIndex);
+      }
+      inputRef.current && inputRef.current.blur();
```
cea2aj, line 51: nit: can we use optional chaining here such as `inputRef.curent?.blur()`

cea2aj:
This approach looks great!

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,122 @@
...
+  useRootClose(containerRef, () => {
+    decoratedToggle(false);
+  }, { disabled: !isOpen });
+
+  useGlobalListener('keydown', e => {
+    if (!isOpen) {
+      return;
+    }
+    if (e.key === 'ArrowDown') {
+      e.preventDefault();
```
nmanu1, line 80: I think Rose wanted the default up/down behavior to be disabled even when there are no dropdown options

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,122 @@
...
+      const updatedFocusedIndex = focusedIndex + 1;
+      setFocusedValue(null);
```
nmanu1, line 82: for my own understanding, why is this set to null each time?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,122 @@
...
+        <InputContext.Provider value={inputContext}>
+          <FocusContext.Provider value={focusContext}>
+            {children}
+          </FocusContext.Provider>
+        </InputContext.Provider>
+      </DropdownContext.Provider>
+
+      <ScreenReader
+        announcementKey={value}
+        announcementText={screenReaderText}
```
nmanu1, line 116: when the dropdown is closed, the sr announcement text should be removed. otherwise it can show up when mousing over with a screen reader and makes it cluttered

src/components/Dropdown/Dropdown.tsx
oshi97: it's not necessary!

src/components/Dropdown/Dropdown.tsx
oshi97: updated!

src/components/Dropdown/Dropdown.tsx
oshi97: we still need to setFocusedValue when returning to the "-1" index which sets the input's value back to the lastTypedOrSubmittedValue

src/components/Dropdown/Dropdown.tsx
oshi97: the ScreenReader now doesn't render is the dropdown is closed, which I think has the right behavior?

src/components/Dropdown/DropdownInput.tsx
oshi97: updated!

src/components/Dropdown/Dropdown.tsx
oshi97: it's unneeded!

src/components/Dropdown/DropdownInput.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const handleClick = () => {
+    decoratedToggle && decoratedToggle(true);
+  };
+
+  const handleChange = (e: ChangeEvent<HTMLInputElement>) => {
+    decoratedToggle && decoratedToggle(true);
+    setFocusedIndex && setFocusedIndex(-1);
+    setLastTypedOrSubmittedValue && setLastTypedOrSubmittedValue(e.target.value);
+    setValue && setValue(e.target.value);
+    onChange && onChange(e.target.value);
```
nmanu1, line 40: nit: could we also use optional chaining for these (e.g. `onChange?.(e.target.value)`) and the other function calls scattered throughout?

src/components/Dropdown/Dropdown.tsx
nmanu1: it would probably be good to keep the sr instructions even when closed, because it's used for the `aria-describedby` of the input element. also, I think it should use `_isOpen`, not `isOpen`, to add the announcement text because we still want announcements when the dropdown is closed because there are no results, just not if the dropdown is escaped/clicked out of

src/components/Dropdown/DropdownItem.tsx
```diff
@@ -0,0 +1,51 @@
...
+    if (needsFocusedValueUpdate) {
+      setFocusedValue && setFocusedValue(value);
+      setValue && setValue(value);
+    }
+  }, [needsFocusedValueUpdate, setFocusedValue, value, setValue])
+
+  const handleClick = () => {
+    decoratedToggle && decoratedToggle(false);
+    setValue && setValue(value);
+    onSelect && onSelect(value);
```
nmanu1, line 36: does it need to set the `lastTypedOrSubmittedValue` here?

src/components/Dropdown/DropdownInput.tsx
oshi97: updated! I'm normally a little worried about throwing an error and crashing the whole app but here I think we shouldn't run into any issues doing so

oshi97: I don't think we want to provide defaults since that might cover up developer errors

src/components/Dropdown/Dropdown.tsx
oshi97: updated!

src/components/Dropdown/DropdownInput.tsx
```diff
@@ -0,0 +1,74 @@
...
+  } = props;
+
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const { decoratedToggle, onSelect, screenReaderUUID } = useContext(DropdownContext) || {};
+  const { value = '', setValue, setLastTypedOrSubmittedValue } = useContext(InputContext) || {};
+  const { focusedIndex = -1, setFocusedIndex } = useContext(FocusContext) || {};
+
+  const handleClick = useCallback(() => {
+    decoratedToggle && decoratedToggle(true);
```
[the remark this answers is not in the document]
oshi97: removed the decoration, this component deals with low level details by design and trying to half-heartedly skirt around them probably just makes things more confusing

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,113 @@
...
+    onSelect,
+    as: ContainerElementType = 'div',
+    className,
+    activeClassName
+  } = props;
+  const containerRef = useRef<HTMLElement>(null);
+
+  const [value, setValue] = useState(initialValue ?? '');
+  const [lastTypedOrSubmittedValue, setLastTypedOrSubmittedValue] = useState(initialValue ?? '');
+  const inputContext: InputContextType = useMemo(() => ({
```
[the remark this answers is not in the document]
oshi97: factored them to the bottom of the file! Since these methods use react hooks, they are react hooks, and react requires all react hooks to start with the "use" prefix

src/components/Dropdown/DropdownContext.ts
```diff
@@ -0,0 +1,13 @@
+import { createContext } from 'react';
+
+/**
+ * The Context responsibile for the Dropdown state.
+ */
+export type DropdownContextType = {
```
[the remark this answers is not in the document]
oshi97: It's for the state of the Dropdown as a whole, it's not really specific to the menu. Kind of just for things that don't fit into `InputContext` or `FocusContext`. Originally I tried to split up the contexts into a way that would reduce renders, so that say the `DropdownInput` would only use the `InputContext` and also `toggleDropdown` from the `DropdownContext`, and since `toggleDropdown` would have a stable reference due to it coming from `React.useState` I could wrap usages of it inside `DropdownInput` within `useCallback/Memo`. BUT as I had to add more and more interactions it just ended up with pretty much every component using most bits of all 3 contexts.

There's probably some argument for just mushing all 3 contexts into a single contexts for simplicity, though I kind of like splitting them up even if `DropdownContext` is kind of just the arbitrary context that has all the bits that don't fit into `InputContext` or `FocusContext`

src/components/Dropdown/Dropdown.tsx
oshi97: I removed the baked in behavior to have `isOpen` be false when numItems < 0, and renamed it to `isActive`

src/components/Dropdown/DropdownItem.tsx
oshi97: it technically doesn't have any drawbacks right now and things work as expected, but I think it should

src/components/Dropdown/DropdownInput.tsx
oshi97: yup! I originally tried to do `onChange?(e.target.value)` but that's not the right syntax

src/components/Dropdown/DropdownItem.tsx
oshi97: yes, updated!

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,113 @@
...
+    value,
+    setValue,
+    lastTypedOrSubmittedValue,
```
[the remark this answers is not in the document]
oshi97: I tried to think of a shorter name but I had a hard time doing so. I didn't like "latestUserInput", which is what we use in InputDropdown right now, or anything similar to that, since it doesn't make it clear that this value is only set when the user types or submits the query, and not when an autocomplete option is focused using the arrow keys - I would have expected this value to only reflect what the user manually types.

I didn't really like something like "baseQuery" either since that doesn't feel descriptive. The purpose of this variable is to be the hidden "-1" indexed value when a user navigates using the autocomplete results with the arrow key, so they can return to whatever original query they had. Would something like "originalQuery" work?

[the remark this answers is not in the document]
yen-tt: I think lastTypedOrSubmittedValue described what the variable is clearly so I don't mind keeping it, but originalQuery or originalUserQuery works too.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,140 @@
...
+  }, { disabled: !isActive });
+
+  useGlobalListener('keydown', e => {
+    if (e.key === 'ArrowDown' || e.key === 'ArrowUp') {
+      e.preventDefault();
+    }
+    if (!isActive) {
+      return;
+    }
+    if (e.key === 'ArrowDown') {
```
yen-tt, line 58: Will we still have DropdownSection or would it now be a list of DropdownItem to navigate?

src/components/Dropdown/Dropdown.tsx
nmanu1: oh ok, so the dropdown being hidden because of no autocomplete options will be handled in some other way, not in `isActive`?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,140 @@
...
+      const updatedFocusedIndex = focusedIndex + 1;
+      if (updatedFocusedIndex >= numItems) {
+        setFocusedIndex(-1);
+        setFocusedValue(lastTypedOrSubmittedValue);
+        setValue(lastTypedOrSubmittedValue);
+      } else {
+        setFocusedIndex(updatedFocusedIndex);
+      }
+    } else if (e.key === 'ArrowUp') {
```
yen-tt, line 67: What's the behavior for `else if (updatedFocusedIndex < -1)`, do we need it? or can this be:
```
if (updatedFocusedIndex < 0) {
  setFocusedIndex(updatedFocusedIndex);
  setFocusedValue(lastTypedOrSubmittedValue);
  setValue(lastTypedOrSubmittedValue);
} else {
  setFocusedIndex(updatedFocusedIndex);
}
```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,140 @@
...
+    focusedValue,
+    setFocusedValue
+  };
+}
+
+function useDropdownContextInstance(
+  value: string,
+  screenReaderUUID: string,
+  onToggle?: (isActive?: boolean, value?: string) => void,
+  onSelect?: (value?: string, index?: number) => void,
```
yen-tt, line 127: do we need to pass in screenReaderUUID and onSelect if it just return it back?

src/components/Dropdown/Dropdown.tsx
oshi97: DropdownSection will be gone after the refactor

src/components/Dropdown/Dropdown.tsx
oshi97: this is for looping around the autocomplete options

src/components/Dropdown/Dropdown.tsx
oshi97: right

oshi97: I think that makes less assumptions, the fact that we need to make a distinction between _isOpen and isOpen within the component itself already is a code smell

src/components/Dropdown/Dropdown.tsx
yen-tt: oh we didn't have that before, I assume this is product request. nvm then!

oshi97: Yup! Nidhi talked to Rose about what arrow key interactions we want

yen-tt (approved):
beside possibly changing the name `lastTypedOrSubmittedValue ` (I'm fine with that or with 'originalQuery'), this looks great!

PR #46, opened 2022-01-19
Visual Autocomplete styling tweaks

cea2aj:
Tweak the VA styling in response to updated mocks

Changes include:
- Update event and FAQ card styling
- Update hover state and padding for autocomplete options
- Update vertical links styling
- Swap highlighting so that the highlighted characters are normal font weight and the nonHighlighted words are bold. Recent searches are left at normal font weight because currently we don't have highlighting for them (deferring to a future item)

J=1791
TEST=visual

Visually compare test site to mocks


src/components/utils/renderHighlightedValue.tsx
```diff
@@ -1,11 +1,13 @@
 import { HighlightedValue } from "@yext/answers-headless-react";
 
 const defaultCssClasses: HighlightedValueCssClasses = {
-  highlighted: 'font-bold'
+  highlighted: 'font-normal',
+  nonHighlighted: 'font-semibold'
```
yen-tt, line 5: I think you meant to swap these two? highlighted should be semibold

src/components/SearchBar.tsx
```diff
@@ -25,7 +25,7 @@ export const builtInCssClasses: SearchBarCssClasses = {
   inputDropdownContainer___active: 'shadow-lg',
   inputElement: 'outline-none flex-grow border-none h-full pl-0.5 pr-2',
   logoContainer: 'w-7 mx-2.5 my-2',
-  optionContainer: 'flex items-stretch mb-3 mx-3.5 cursor-pointer',
+  optionContainer: 'flex items-stretch py-1.5 px-3.5 cursor-pointer hover:bg-gray-100',
```
yen-tt, line 28: I think this could be a problem in terms of option highlighting when you switch between keyboard navigation and mouse hovering. For example, if you navigate with keyboard down to an option, then your mouse is hover over another, there would be two option 'highlighted' in gray

src/components/VisualAutocomplete/SampleVisualSearchBar.tsx
```diff
@@ -46,11 +51,26 @@ interface FaqData {
...
-    <div tabIndex={0} className='flex flex-col mx-4 mb-3 rounded-md'>
-      <div className='text-gray-900 text-lg font-medium pb-1'>{faqData.question}</div>
-      <div>{faqData.answer}</div>
+    <div className='h-px bg-gray-200 mt-1 mb-4 mx-2.5'></div>
+    <div tabIndex={0} className='flex flex-row mx-4 mb-3 rounded-md'>
+      <div className='w-6 mr-3 mt-1'>
+        <FAQIcon />
+      </div>
+      <div>
+        <div className='text-gray-800 text-lg font-semibold pb-1'>{renderHighlightedValue({ value: faqData.question, matchedSubstrings }, { highlighted: 'font-normal' })}</div>
```
yen-tt, line 71: probably `highlighted: 'font-semibold'` and `font-normal` for the div instead?

src/components/VisualAutocomplete/SampleVisualSearchBar.tsx
```diff
@@ -46,11 +51,26 @@ interface FaqData {
...
+
   return (<div key={faqData.question}>
-    <div className='h-px bg-gray-200 mt-1 mb-3 mx-2.5'></div>
-    <div tabIndex={0} className='flex flex-col mx-4 mb-3 rounded-md'>
-      <div className='text-gray-900 text-lg font-medium pb-1'>{faqData.question}</div>
-      <div>{faqData.answer}</div>
+    <div className='h-px bg-gray-200 mt-1 mb-4 mx-2.5'></div>
+    <div tabIndex={0} className='flex flex-row mx-4 mb-3 rounded-md'>
+      <div className='w-6 mr-3 mt-1'>
+        <FAQIcon />
```
yen-tt, line 68: can we have `<FAQIcon className='w-6 mr-3 mt-1' />` instead of wrapping it in another div?

src/components/utils/renderHighlightedValue.tsx
cea2aj: Michael actually updated the mocks so that the highlighted text is normal font weight and the nonHighlighted text is semibold

src/components/SearchBar.tsx
cea2aj: I'll check with UX if they have any suggestions here

src/components/VisualAutocomplete/SampleVisualSearchBar.tsx
cea2aj: per the mocks, they are supposed to be inverted

src/components/VisualAutocomplete/SampleVisualSearchBar.tsx
cea2aj: sure, that works

src/components/SearchBar.tsx
cea2aj: UX is okay with having highlighting when hovered and also with keyboard navigation. Apparently lots of other sites have this behavior

yen-tt: gotcha. Sounds good!

yen-tt (approved):

PR #48, opened 2022-01-20
Scaffold Wrapper Component

yen-tt:
This pr added changes so the repo can create a react site given a JSON Schema that we expect PageBuilder 2.0 would provide. 

- All page builder related files are in pageBuilder/ folder.
- the expected JSON schema is described by interface in `AnswersAppConfig.ts` and an example of the config is in `pageBuilderAnswersAppConfig.ts`
  - Note: This currently does NOT support everything that the components in the repo can do. Also can't specify the Page type to use yet.
- the wrapper component `PageBuilderApp`, takes in a `config` field where the JSON would get pass in. It would construct the routes and pages based on that config
   - it will use the two page templates `UniversalStandardPage` and `VerticalStandardPage` 
- `AnswersAppContext` is used to provide the shared config for all children components of the app. This include the universal and vertical configurations necessary to construct navigation and vertical links in Navigation component and VisualSearchBar component. 
- useEntityPreview originally pull headless config from a static file. This is not applicable for component library or pageBuilder. Updated VisualSearchbar and useEntityPreview to require a new searcher for useEntityPreview. The searcher can be created using `useEntityPreviewSearcher`

J=SLAP-1789
TEST=manual

see that localhost site works as expected with App and PageBuilderApp.


src/config/pageBuilderAnswersAppConfig.ts
```diff
@@ -0,0 +1,49 @@
...
+  },
+  verticals: {
+    faqs: {
+      label: 'FAQS',
+      path: '/faqs',
+      cardConfig: {
+        cardName: "STANDARD"
+      }
+    },
+    events: {
```
oshi97, line 26: does this let us delete universalResultsConfig.ts?

src/models/cardComponent.ts
```diff
@@ -1,11 +1,18 @@
...
 
 /**
  * The config types for each supported card.
  */
 export type CardConfigTypes = StandardCardConfig;
 
+/**
+ * Map card name to the corresponding React component.
+ */
+export const CardNameToComponentMapping: Record<string, CardComponent> = {
```
oshi97, line 12: I assume product wanted this mapping, but the only reason I can see for wanting this kind of mapping is wanting to use a pure JSON schema instead of a JS file. Was there any talks of allowing a JS file instead? We ran into some flexibility issues with the theme by restricting things to JSON, and did not really get many benefits from it.

src/pages/StandardLayout.tsx
```diff
@@ -29,7 +29,7 @@ const StandardLayout: LayoutComponent = ({ page }) => {
         />
         : <SampleVisualSearchBar />
       }
-      <Navigation links={navLinks} />
+      {navLinks && <Navigation links={navLinks} />}
```
oshi97, line 32: is navLinks ever falsy?

yen-tt: nope! I was trying out another approach that would make it falsy but the current way shouldn't, will remove.

src/models/cardComponent.ts
yen-tt: From Rose's config draft, it looks like product wanted this mapping, similar to what theme has. I'm waiting to confirm with tom if this restriction is needed. But I agree! A js file would make things easier

src/config/pageBuilderAnswersAppConfig.ts
yen-tt: sort of, the App for starter site demo requires routeConfig to point to different vertical pages (vs PageBuilderApp that currently only support StandardPage template), and routeConfig requires universalResultsConfig with some configs that can't be directly pull from the JSON schema. Originally, I was thinking we can get rid of routeConfig and universalResultsConfig once we add the ability for custom page templates. But for now, I could add a bit of logic in routeConfig to sort out the field mapping and get rid of universalResultsConfig.

oshi97: sgtm!

src/config/starterAnswersAppConfig.ts
```diff
@@ -0,0 +1,32 @@
+import AnswersAppConfig from "./AnswersAppConfig";
```
cea2aj, line 1: Can you update the README to describe the config changes?

src/index.tsx
```diff
@@ -4,10 +4,13 @@ import './tailwind.css';
 import './sass/index.scss';
 import App from './App';
 import registerServiceWorker from './registerServiceWorker';
+// import PageBuilderApp from './PageBuilderApp';
```
cea2aj, line 7: Is this left here on purpose for testing?

yen-tt: yup

src/config/AnswersAppConfig.ts
```diff
@@ -0,0 +1,53 @@
+import { SortType, Direction } from "@yext/answers-headless-react";
```
cea2aj, line 1: What do you think of defining this outside of the src/config folder? In my mind the config folder should only contain config. Interfaces seem a little out of place there

src/config/pageBuilderAnswersAppConfig.ts
```diff
@@ -0,0 +1,49 @@
+import AnswersAppConfig from "./AnswersAppConfig";
```
cea2aj, line 1: Can we move this to a different config folder as well? I can imagine someone forking the sample app, looking at the config folder, and being confused as to why changing this doesn't affect anything

src/config/starterAnswersAppConfig.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const starterAnswersAppConfig: AnswersAppConfig = {
```
cea2aj, line 3: this interface looks great!

cea2aj: Keep in mind that the hitchhikers have some documentation which describes how to use the sample app and explains our previous config (answersHeadlessConfig, routeConfig, and universalResultsConfig), so we'll need to make sure they update that when we release this version

src/pages/UniversalSearchPage.tsx
```diff
@@ -1,13 +1,14 @@
...
 import DirectAnswer from '../components/DirectAnswer';
-import { UniversalResultsConfig } from '../config/universalResultsConfig';
 import SpellCheck from '../components/SpellCheck';
 import usePageSetupEffect from '../hooks/usePageSetupEffect';
 
 const universalResultsFilterConfig = {
   show: true
 };
 
+export type UniversalResultsConfig = Record<string, VerticalConfig>;
```
cea2aj, line 10: What do you think of having the UniversalResults component define and export this?

src/index.tsx
cea2aj: That makes sense, however is there another spot we could put this so that we don't confuse regular users of the library which don't know what PageBuilder is? Maybe we could have a `src/pagebuilder` folder for some of the pagebuilder work?

src/config/AnswersAppConfig.ts
yen-tt: moved to model/ folder

src/config/pageBuilderAnswersAppConfig.ts
yen-tt: moved to pagebuilder/ folder!

src/index.tsx
yen-tt: made another index.js to build page builder app! This would require a small change in `path.js`, as noted in the comment I added in pagebuilder/index.js. If we want, we can make a script to streamline this, but that could be in another item.

src/config/starterAnswersAppConfig.ts
yen-tt: I decided to keep all the original configs for the starter app as is. If we want to consolidate that into one config, let's make a new item for that.

src/config/starterAnswersAppConfig.ts
yen-tt: I'm keeping the starter app config as is for now so no documentation change is needed

src/hooks/useEntityPreviews.tsx
```diff
@@ -15,18 +21,17 @@ type ExecuteEntityPreviewsQuery = (query: string, universalLimit: UniversalLimit
...
+export function useEntityPreviews(headlessConfig: HeadlessConfig, debounceTime: number):[ EntityPreviewsState, ExecuteEntityPreviewsQuery ] {
   const headlessRef = useRef<AnswersHeadless>();
-  if (!headlessRef.current) {
-    headlessRef.current = provideAnswersHeadless({
-      ...answersHeadlessConfig,
-      headlessId
-    });
-    headlessRef.current.setQuerySource(QuerySource.Autocomplete);
-  }
+  useEffect(() => {
```
yen-tt, line 29: There was a component update/render error due to state update order conflict between instantiating a new answers headless (trigger state update from replaceReducer()) and the setState() from usePageSetupEffect. This useEffect call ensures usePageSetupEffect changes is reflected in state before combining reducers and updating store's state again.

src/hooks/useEntityPreviews.tsx
```diff
@@ -15,18 +21,17 @@ type ExecuteEntityPreviewsQuery = (query: string, universalLimit: UniversalLimit
  * useEntityPreviews provides state surrounding the visual entities portion of visual autocomplete,
  * which performs debounced universal searches.
  * 
- * @param headlessId a unique id for the new headless instance that will be created by the hook
+ * @param headlessConfig config for the new headless instance that will be created by the hook
  * @param debounceTime the time in milliseconds to debounce the universal search request
  */
-export function useEntityPreviews(headlessId: string, debounceTime: number):[ EntityPreviewsState, ExecuteEntityPreviewsQuery ] {
+export function useEntityPreviews(headlessConfig: HeadlessConfig, debounceTime: number):[ EntityPreviewsState, ExecuteEntityPreviewsQuery ] {
```
yen-tt, line 27: I changed this to pass in headlessConfig in order to instantiate a new headless using the same config as the top level headlessContextProvider. Another approach, after discussing with tom, was to make a getter method in answers-headless repo to return a config given a headless id. However, that still require the original id to be pass into useEntityPreviews to get the config and override it with the new id. So, I figured we might as well just pass in the original config. Let me know if that's another preferred approach!

src/components/UniversalResults.tsx
```diff
@@ -24,6 +24,8 @@ export interface VerticalConfig {
   viewAllButton?: boolean
 }
 
+export type UniversalResultsConfig = Record<string, VerticalConfig>;
```
cea2aj, line 27: Can you update the props to use this config?

src/components/VisualAutocomplete/SampleVisualSearchBar.tsx
```diff
@@ -1,21 +1,23 @@
...
 import { ReactComponent as FAQIcon } from '../../icons/faq.svg';
 import renderHighlightedValue from '../utils/renderHighlightedValue';
 
 /**
  * This is an example of how to use the VisualSearchBar component.
  */
-export default function SampleVisualSearchBar() {
+export default function SampleVisualSearchBar({ verticalKeyToLabelMap, entityPreviewHeadlessConfig }: { 
+  verticalKeyToLabelMap?: Record<string, string>, 
+  entityPreviewHeadlessConfig: HeadlessConfig
```
tmeyer2115, line 13: Instead of passing a `HeadlessConfig` all the way down to the `useEntityPreview` hook, we could take in the `HeadlessConfig` here and make a new, name-spaced searcher. That searcher would then be passed as props to the `VisualSearchBar` Component. I think, conceptually at least, that might be cleaner. What the `VisualSearchBar` needs is a new searcher, not a `HeadlessConfig` instance.

src/pagebuilder/AnswersAppConfig.ts
```diff
@@ -0,0 +1,54 @@
+import { SortType, Direction } from "@yext/answers-headless-react";
+import { AnswersHeadlessProvider } from '@yext/answers-headless-react';
+
+export type HeadlessProviderConfig = Parameters<typeof AnswersHeadlessProvider>[0];
```
tmeyer2115, line 4: For my own edification, what does `Parameters<typeof AnswersHeadlessProvider>` get us? Have you considered expressing this type as a function of `AnswersConfigWithApiKey` with certain things (endpoints, visitor, additionalQueryParams) removed?

src/pagebuilder/AnswersAppConfig.ts
```diff
@@ -0,0 +1,54 @@
...
+  type: SortType,
+  field?: string,
+  direction?: Direction
+}
+
+interface SearchBarConfig {    
+  placeholder?: string 
+}
+
+export interface VerticalPageConfig {
```
tmeyer2115, line 21: Could we break this file up? It would contain the top-level structure of the app config. But, some of the constituent interfaces could be moved to their own files.

src/pagebuilder/AnswersAppContext.tsx
```diff
@@ -0,0 +1,24 @@
+
+import { createContext, useContext } from 'react';
+import AnswersAppConfig from './AnswersAppConfig';
+
+export const AnswersAppContext = createContext<AnswersAppConfig | undefined>(undefined);
+
+export function AnswersAppContextProvider(
```
tmeyer2115, line 7: JSDocs

src/pagebuilder/PageBuilderApp.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { AnswersHeadlessProvider } from '@yext/answers-headless-react';
+import { PageViewContextProvider } from '../context/PageViewContext';
+import AnswersAppConfig from './AnswersAppConfig';
+import VerticalStandardPage from './templates/VerticalStandardPage';
+import { AnswersAppContextProvider } from './AnswersAppContext';
+import { UniversalResultsConfig } from '../components/UniversalResults';
+import UniversalStandardPage from './templates/UniversalStandardPage';
+import { CardNameToComponentMap } from './templates/componentMapping';
+
+export default function PageBuilderApp({ config }: { config: AnswersAppConfig }) {
```
tmeyer2115, line 11: Separate `Props` interface

src/pagebuilder/PageBuilderApp.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const verticalRoutesConfig: RouteData[] = [];
+  const universalResultsConfig : UniversalResultsConfig = {};
+
+  const navLinks = [
```
tmeyer2115, line 15: Can we extract this to its own method?

src/pagebuilder/PageBuilderApp.tsx
```diff
@@ -0,0 +1,56 @@
...
+      label: config.universal.label || 'All'
+    },
+    ...Object.entries(config.verticals).map(([verticalKey, config]) => ({
+      to: config.path || `/${verticalKey}`,
+      label: config.label || verticalKey
+    }))
+  ];
+
+  Object.entries(config.verticals).forEach(([key, config]) => {
+    const CardComponent = CardNameToComponentMap[config.cardConfig?.cardName || 'STANDARD'];
```
tmeyer2115, line 27: Cool, so it seems like you already created the card registry that we needed for SLAP-1807

src/pagebuilder/PageBuilderApp.tsx
```diff
@@ -0,0 +1,56 @@
...
+      to: '/',
+      label: config.universal.label || 'All'
+    },
+    ...Object.entries(config.verticals).map(([verticalKey, config]) => ({
+      to: config.path || `/${verticalKey}`,
+      label: config.label || verticalKey
+    }))
+  ];
+
+  Object.entries(config.verticals).forEach(([key, config]) => {
```
tmeyer2115, line 26: Separate method for this too.

src/pagebuilder/PageBuilderApp.tsx
```diff
@@ -0,0 +1,56 @@
...
+    const CardComponent = CardNameToComponentMap[config.cardConfig?.cardName || 'STANDARD'];
+    universalResultsConfig[key] = { 
+      label: config.label,
```
tmeyer2115, line 29: Do we need the `config.label || key` default here?

src/pagebuilder/AnswersAppConfig.ts
yen-tt: That give us the params use be AnswersHeadlessProvider (HeadlessConfig & {verticalKey, sessionTrackingEnabled, etc. }). Ideally, in answers-headless-react repo, we should export this type so it can be use in the starter repo instead of extracting the type with this typescript utility. We can make an item for this, and the type define here can be remove. what do you think?

src/pages/StandardLayout.tsx
```diff
@@ -16,18 +18,30 @@ const navLinks = [
...
+Object.entries(universalResultsConfig)
+  .forEach(([key, config]) => { verticalKeyToLabelMap[key] = config.label ?? key })
+
+
 /**
  * A LayoutComponent that provides a SearchBar and Navigation tabs to a given page.
  */
 const StandardLayout: LayoutComponent = ({ page }) => {
   const isVertical = useAnswersState(s => s.meta.searchType) === SearchTypeEnum.Vertical;
+  const entityPreviewSearcher = useEntityPreviewSearcher({
```
oshi97, line 31: for my edification why did we pull this out?

yen-tt: I originally had the headless config pass into useEntityPreviews. But tom pointed out that it would make more sense in visualSearchBar interface to require a searcher for entity preview instead of a config

src/pages/StandardLayout.tsx
```diff
@@ -16,18 +18,30 @@ const navLinks = [
   }))
 ]
 
+const verticalKeyToLabelMap: Record<string, string> = {};
+Object.entries(universalResultsConfig)
+  .forEach(([key, config]) => { verticalKeyToLabelMap[key] = config.label ?? key })
```
cea2aj, line 23: nit: not super important but this could be done with the reduce function instead

cea2aj (approved):
LGTM!

src/hooks/useEntityPreviewSearcher.tsx
```diff
@@ -0,0 +1,25 @@
...
+  AnswersHeadless
+} from "@yext/answers-headless-react";
+import { useEffect, useState } from "react";
+
+/**
+ * Create a new AnswersHeadless searcher for Entity Previews.
+ *
+ * @param searcherConfig - config for the new AnswersHeadless instance
+ */
+export default function useEntityPreviewSearcher(searcherConfig: HeadlessConfig): AnswersHeadless | undefined {
```
tmeyer2115, line 14: What benefit do we get by using the `useEffect` hook here and setting the generated searcher in state? Could we not make this a simple factory method without any of these hooks?

src/pages/StandardLayout.tsx
yen-tt: sure thing

src/hooks/useEntityPreviewSearcher.tsx
yen-tt: without the useEffect hook, we have an error due to a state update conflict between creating a new answers headless instance and the setState call inside usePageSetupEffect's useLayoutEffect. This occurs when you switch from a vertical page to a universal page, during the render stage. Creating a new answers headless would call replaceReducer() which trigger a state update to merge existing headless state tree with the new one, and usePageSetupEffect would also attempt to update state and trigger render of components from previous page.

tmeyer2115 (approved):

PR #51, opened 2022-01-21
fix renderHighlighedValue styling

oshi97:
It appears that the default css classes are flipped.

TEST=manual

see that highlighted values dont have flipped styling


nmanu1:
I think this is intentional based on the latest UX mocks Michael made

PR #52, opened 2022-01-24
Consolidate SearchBar and VisualSearchBar using Dropdown

oshi97:
This commit consolidates the two SearchBar's we have, by using the newly added Dropdown component.

As part of this, Dropdown was updated to no longer need a `numItems` prop on the main component,
and `index` props for `DropdownItem`s.
This was done by auto-magically recursing through child components.
I don't believe this is an inherently bad pattern, it just needs to be applied carefully.
One difference between this pattern's usage here, vs how it was used for the original InputDropdown, is that we don't expose the injected props to the user. Instead, there is an external facing component, DropdownItem, and an internal DropdownItemWithIndex that is used internally by Dropdown.

This change was necessary due to the complexity of the DropdownItems necessary for VisualAutocomplete.
Making user's manually specify them is fine for simple cases, and is almost certainly more performant, but for more complex cases with multiple arrays of data, it quickly becomes hard to juggle between which item is which index, and oddly restrictive in terms code formatting.
I did not notice any performance regressions from recursing through props.children.

In order to support verticalLinks, an arbitrary "metadata" attribute was added to DropdownItem.
I initially attempted to use generics for this, but found them unwieldy and not terribly helpful.
I don't foresee us needing to store massive amounts of metadata.

I also removed the ability to specify your own headlessId. It doesn't seem particularly useful, and clutters up the props, especially since it's completely unused for the regular SearchBar case.

Removed the `VisualAutocomplete/` folder, moved `EntityPreviews.tsx` and `SampleVisualSearchBar.tsx` down a level.
`filteredRecentSearches` is no longer its own state variable because it's not necessary.

J=SLAP-1777
TEST=manual

tested vertical links, both by clicking on them with the mouse, and focusing them with the arrow keys and hitting enter (tested on the slanswers experience since the answers-js-docs does not have sufficient test data)

vertical links do not appear for vertical search pages

tested selecting recent searches, and that they have the correct styling when focused with the arrow keys
recent searches do not show up on vertical search pages
recent searches are still filtered by the current query and levenshtein distance


src/components/SearchBar.tsx
```diff
@@ -56,77 +90,234 @@ interface Props {
...
-      value: result.value,
-      onSelect: () => {
-        autocompletePromiseRef.current = undefined;
-        answersActions.setQuery(result.value);
-        executeQuery();
-      },
-      display: renderAutocompleteResult(result, cssClasses, MagnifyingGlassIcon, `autocomplete option: ${result.value}`)
+  const [executeQueryWithNearMeHandling, autocompletePromiseRef] = useSearchWithNearMeHandling(answersActions, geolocationOptions);
+  const [recentSearches, setRecentSearch, clearRecentSearches] = useRecentSearches(recentSearchesLimit);
+  const filteredRecentSearches = recentSearches?.filter(search =>
```
nmanu1, line 118: does computing the `filteredRecentSearches` each render cause the displayed recent searches to change when navigating the dropdown with arrow keys?

oshi97: No it doesn't do that - is that the behavior we want? it doesn't look like we had that behavior before, and also seems a little odd?

nmanu1: yeah, we don't want it to do that. I was just double checking because I think I remember having that be an issue when I first added the filtering

src/components/SearchBar.tsx
```diff
@@ -56,77 +90,238 @@ interface Props {
...
+  }, [clearRecentSearches, hideRecentSearches])
 
-  const screenReaderText = processTranslation({
-    phrase: `${options.length} autocomplete option found.`,
-    pluralForm: `${options.length} autocomplete options found.`,
-    count: options.length
-  });
+  function executeQuery() {
+    if (!hideRecentSearches) {
+      const input = answersActions.state.query.input;
```
cea2aj, line 130: is there a reason for accessing the input this way as opposed to using the query from `useAnswersState` above?

src/components/SearchBar.tsx
```diff
@@ -56,77 +90,238 @@ interface Props {
...
+      input && setRecentSearch(input);
+    }
+    executeQueryWithNearMeHandling();
+  }
+
+  const [entityPreviewsState, executeEntityPreviewsQuery] = useEntityPreviews(entityPreviewsDebouncingTime);
+  const { verticalResultsArray, isLoading: entityPreviewsLoading } = entityPreviewsState;
+  const entityPreviews = renderEntityPreviews && renderEntityPreviews(entityPreviewsLoading, verticalResultsArray);
+  function updateEntityPreviews(query: string) {
```
cea2aj, line 139: I'm just curious what your thoughts were on making query a param rather than getting it from answers state. I suppose that would mean you would need to call `answersActions.setQuery()` before calling `updateEntityPreviewes`. By using a param you can avoid issues with the order things are called

src/components/SearchBar.tsx
```diff
@@ -56,77 +90,238 @@ interface Props {
...
+              { value: `in ${verticalKeyToLabel ? verticalKeyToLabel(verticalKey) : verticalKey}` },
+              { ...cssClasses, option: cssClasses.verticalLink }
+            )}
+          </DropdownItem>
+        ))}
+      </Fragment>
+    ))
+  }
+
+  const hasItems = !!(autocompleteResponse?.results.length  || (!isVertical && filteredRecentSearches?.length));
```
cea2aj, line 240: I don't think product has asked for this yet, but I don't really see a reason to not allow recent searches on verticals

src/components/SearchBar.tsx
oshi97: updateEntityPreviews uses the separate headless instance, so we can't just get it from answers state. This is also how things are done in the current VisualSearchBar.

src/components/SearchBar.tsx
oshi97: I wanted to avoid scope creep

src/components/SearchBar.tsx
oshi97: This function was copied from the old VisualSearchBar, but I can update.

src/components/Dropdown/DropdownItem.tsx
```diff
@@ -4,47 +4,67 @@ import { useFocusContext } from './FocusContext';
...
 
-export default function DropdownItem(props: PropsWithChildren<{
+type DropdownItemProps = PropsWithChildren<{
   value: string,
-  index: number,
   className?: string,
-  focusedClassName?: string
-}>) {
+  focusedClassName?: string,
+  metadata?: Record<string, unknown> | undefined
```
cea2aj, line 11: When I think of metadata, I think of data that isn't critical to the functionality of application, however here it's being used for vertical links which to me is more significant. What do you think about renaming to 'itemData' or just 'data'?

src/components/SearchBar.tsx
oshi97: Also probably the same issue noted above with using the query from the previous render and not the current one (though I haven't tested)

src/components/SearchBar.tsx
oshi97: Actually, it looks like if you just use the `query` state variable, the recent search will be set to the query from the previous render, not the current one. This is likely due to how `query` will be snapshotted to a fixed value per render, while `answersActions.state.query.input` will always be the most up to date query

src/components/Dropdown/DropdownItem.tsx
oshi97: renamed to itemData!

src/components/SearchBar.tsx
oshi97: This is a little different than ordering issues so much as render cycle issues

src/components/SearchBar.tsx
oshi97: I think using `answersActions.state.query.input` could actually run into ordering issues, though.

src/components/Dropdown/DropdownInput.tsx
```diff
@@ -35,6 +35,7 @@ export default function DropdownInput(props: {
   const handleChange = (e: ChangeEvent<HTMLInputElement>) => {
     toggleDropdown(true);
     setFocusedIndex(-1);
```
yen-tt, line 37: do we also need to call setFocusedValue in handleChange and handleKeyDown?

src/components/SearchBar.tsx
```diff
@@ -56,77 +90,238 @@ interface Props {
...
+    return (
+      <DropdownInput
+        className={cssClasses.inputElement}
         placeholder={placeholder}
-        screenReaderInstructions={SCREENREADER_INSTRUCTIONS}
-        screenReaderText={screenReaderText}
-        onSubmit={executeQuery}
-        onInputChange={value => {
+        onSubmit={(_value, index) => {
+          if (index < 0) {
```
yen-tt, line 166: how come we don't want to call executeQuery on other indexes?

src/components/Dropdown/DropdownItem.tsx
```diff
@@ -4,47 +4,67 @@ import { useFocusContext } from './FocusContext';
...
+    setFocusedItemData,
+    focusedItemData
+  } = useFocusContext();
   const { setValue, setLastTypedOrSubmittedValue } = useInputContext();
 
-  const needsFocusedValueUpdate = focusedIndex === index && focusedValue !== value;
+  const isFocused = focusedIndex === index;
+  const needsFocusedValueUpdate = isFocused && focusedValue !== value;
+  const needsFocusedItemDataUpdate = isFocused && focusedItemData !== itemData;
   useLayoutEffect(() => {
```
yen-tt, line 44: can we call those these setters without useLayoutEffect?

src/components/Dropdown/DropdownInput.tsx
oshi97: nope! that change is caused by DropdownItem

src/components/SearchBar.tsx
oshi97: if there is a focused DropdownItem, hitting enter will cause both onSubmit and onSelect (a prop in the main Dropdown component) to run, causing two queries to be fired. I didn't want to make onSubmit run only when there is no focused DropdownItem, which I feel like is a little confusing but not 100% sure

src/components/Dropdown/DropdownItem.tsx
oshi97: we definitely shouldn't call setters within a render, but I think since I caved and updated Dropdown to recurse through its children this can probably be changed

src/components/SearchBar.tsx
yen-tt: When selecting an option, it seems like the code is making an assumption that it should always submit as well, which I think may not always be the behavior that user want? In dropdownInput, on enter, what do you think about updating it to only call onSubmit (focusedIndex < 0) or onSelect (focusedIndex >= 0) but never both. This way, only one executeQuery call is made, and we also allow user the choice of a behavior where they can select an option and update searchbar or whatever logic they want without necessarily submitting.

src/components/Dropdown/DropdownInput.tsx
oshi97: updated now!

src/components/SearchBar.tsx
oshi97: in my mind whenever you press enter you are submitting something, though maybe we could change the name from onSubmit

oshi97: If we don't change the callback's name I think I would always want to run it when the user hits enter

src/components/SearchBar.tsx
```diff
@@ -30,24 +40,48 @@ export const builtInCssClasses: SearchBarCssClasses = {
...
+  verticalLink: 'ml-12 pl-1 text-gray-500 italic',
+
+  ...AutocompleteResultBuiltInCssClasses
+};
 
-export interface SearchBarCssClasses 
-  extends InputDropdownCssClasses, DropdownSectionCssClasses, AutocompleteResultCssClasses 
-{
+export interface SearchBarCssClasses
+  extends InputDropdownCssClasses, DropdownSectionCssClasses, AutocompleteResultCssClasses {
```
yen-tt, line 54: can we remove DropdownSectionCssClasses?

src/components/SearchBar.tsx
```diff
@@ -56,77 +90,238 @@ interface Props {
...
+    sectionContainer?: string,
+    optionsContainer?: string
+  }
+}>) {
+  return (
+    <DropdownMenu>
+      <div className={cssClasses.divider} />
+      <div className={cssClasses.dropdownContainer}>
+        <div className={cssClasses.sectionContainer}>
+          <div className={cssClasses.optionsContainer}>
```
yen-tt, line 304: do we need all three divs?

oshi97: yes

oshi97: this is what we had before in InputDropdown as well

src/components/SearchBar.tsx
oshi97: removed!

src/components/SearchBar.tsx
yen-tt: I think they can be merge somehow? At least with sectionContainer and optionsContainer, to combine and rename, since it previously is use for individual section, which is no longer the case since this wraps around the entire dropdown menu I think

oshi97: In that case I think we would also want to rethink the injectable css classes for this component, since there would be no difference between dropdownContainer, sectionContainer, and optionsContainer. I think it would be good to cut down on the number of css classes somebody can inject (there's over 20 right now...) but I think that should be its own item? Unless you already have concrete ideas about what the names of the css classes should be?

src/components/SearchBar.tsx
yen-tt: hmm I think having the name onSubmit prop in searchbar do make sense though. If we are keeping this if clause, maybe add a comment about avoiding another search queries? Maybe others have different thoughts on this.

src/components/SearchBar.tsx
cea2aj: I agree it's out of scope

yen-tt: yea we could make another item for this

src/components/SearchBar.tsx
cea2aj: If we want to keep this as-is, I think adding a comment would be helpful. Another approach (potentially a separate item) is to create a debounced `executeQuery` so that we can remove the if statement in the onSubmit callback. The debouncer will prevent 2 queries from being fired, and it's a good thing to have in general

oshi97: I don't think we should add debouncing here, the SDK did that and it masked a lot of dumb code smells

oshi97: Here I think it makes more sense to remove the onSelect handler, and add an onClick handler to DropdownItem (which existed before while prototyping but I removed because I wasn't using it)

oshi97: the onSelect handler is still useful for FilterSearch so I assume it's still useful in general

src/components/Dropdown/Dropdown.tsx
```diff
@@ -109,22 +102,49 @@ function useInputContextInstance(initialValue = ''): InputContextType {
...
+  lastTypedOrSubmittedValue: string,
+  setValue: (newValue: string) => void
+): FocusContextType {
   const [focusedIndex, setFocusedIndex] = useState(-1);
-  const [focusedValue, setFocusedValue] = useState<string>('');
+  const [focusedValue, setFocusedValue] = useState<string | null>(null);
+  const [focusedItemData, setFocusedItemData] = useState<Record<string, unknown> | undefined>(undefined);
+  const numItems = items.length;
+
+  function updateFocusedIndex(updatedFocusedIndex: number, value?: string) {
```
cea2aj, line 115: What do you think of renaming this to `updatedFocusedItem` since it's operating on and affecting more than just the index now

oshi97: updated!

cea2aj (approved):
LGTM!

PR #53, opened 2022-01-26
update FilterSearch to use Dropdown

oshi97:
Added `submitCriteria` prop to `DropdownInput`.
This prevents `DropdownInput` from performing a submission (i.e. triggered by hitting the enter key) if the criteria is not met.

J=SLAP-1838
TEST=manual

side by side comparison of original FilterSearch with current FilterSearch
tested queries
'V', 'Visa', 'Pay', 'Mastercard', 'Drip'
tested that submissions can only be made if an option is focused
arrow keys can loop through options
filters are applied on option selection


src/components/FilterSearch.tsx
```diff
@@ -51,43 +58,85 @@ export default function FilterSearch ({
...
+            answersActions.executeVerticalQuery();
+          }
+        }}
+      >
+        <div className={cssClasses.inputContainer}>
+          <DropdownInput
+            className={cssClasses.inputElement}
+            placeholder='Search here ...'
+            onChange={query => executeFilterSearch(query)}
+            submitCriteria={(_value, index) => index >= 0}
```
nmanu1, line 112: Is there a case in which the value would also be used instead of just the index?

src/components/FilterSearch.tsx
```diff
@@ -51,43 +58,85 @@ export default function FilterSearch ({
...
+          />
+        </div>
+        <DropdownMenu>
+          {hasResults && <>
+            <div className={cssClasses.divider} />
```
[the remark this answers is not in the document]
oshi97: removed!

src/components/FilterSearch.tsx
oshi97: I wanted the hooks to have the same kind of interface of `(value: string, index: number, itemData: FocusedItemData) => ???` but I don't realistically see a use for value. We might want to add some kind of "disabled" status for DropdownItems in the future but I still don't think we would need this

oshi97: it's just index now!

yen-tt (commented):
When I was testing searchbar in this branch. I noticed a couple potential bugs:
  -input doesn't get updated when I arrow down to different options and when I hit enter on one of the options
  -when you click on search button, autocomplete doesn't close

src/components/Dropdown/DropdownInput.tsx
```diff
@@ -44,7 +46,7 @@ export default function DropdownInput(props: {
   };
 
   const handleKeyDown = (e: KeyboardEvent<HTMLInputElement>) => {
-    if (e.key === 'Enter') {
+    if (e.key === 'Enter' && (!submitCriteria || submitCriteria(focusedIndex))) {
```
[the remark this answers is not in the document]
oshi97: no because when the return value of submitCriteria is true we should run a submit

src/components/FilterSearch.tsx
```diff
@@ -33,16 +40,15 @@ export interface FilterSearchProps {
...
   label,
   sectioned,
   searchFields,
   customCssClasses,
   cssCompositionMethod
 }: FilterSearchProps): JSX.Element {
   const answersActions = useAnswersActions();
-  const [input, setInput] = useState('');
-  const selectedFilterOptionRef = useRef<Filter|null>(null);
+  const selectedFilterOptionRef = useRef<Filter | null>(null);
```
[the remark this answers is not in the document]
oshi97: yes that works!

oshi97:
> When I was testing searchbar in this branch. I noticed a couple potential bugs (probably address in another item/pr): -input doesn't get updated when I arrow down to different options and when I hit enter on one of the options -when you click on search button, autocomplete doesn't close

that's my bad these were introduced when I added the `controlledQuery` prop to the Dropdown at the end of the previous PR, I will address them here since I don't think they should be too huge

oshi97:
> When I was testing searchbar in this branch. I noticed a couple potential bugs (probably address in another item/pr): -input doesn't get updated when I arrow down to different options and when I hit enter on one of the options -when you click on search button, autocomplete doesn't close

annd they should be fixed! controlledQuery was added originally to fix an issue with spellcheck where clicking the spellcheck wouldn't update the SearchBar's query - now instead of watching the Dropdown's `value` it watches the `lastTypedOrSubmittedValue` which I think fixes things
to fix the SearchButton not toggling the `Dropdown` I made use of the `DropdownContext` inside a new `DropdownSearchButton` component specified within `SearchBar.tsx`

src/components/SearchBar.tsx
```diff
@@ -333,3 +326,26 @@ function getScreenReaderText(autocompleteOptions = 0, recentSearchesOptions = 0)
...
+  isLoading: boolean,
+  cssClasses: {
+    searchButtonContainer?: string,
+    submitButton?: string
+  }
+}) {
+  const { toggleDropdown } = useDropdownContext();
+  return (
+    <div className={cssClasses.searchButtonContainer}>
+      <SearchButton
```
yen-tt, line 341: I think SearchBar is the only component that uses SearchButton. Can we merge the two components and have it inside SearchBar instead of having both SearchButton component file and DropdownSearchButton in here?

src/components/FilterSearch.tsx
```diff
@@ -1,17 +1,24 @@
-import { useRef, useState } from "react";
-import { useAnswersActions, FilterSearchResponse, SearchParameterField, Filter } from '@yext/answers-headless-react';
-import InputDropdown, { InputDropdownCssClasses } from "./InputDropdown";
```
yen-tt, line 3: I think we can delete InputDropdown and DropdownSection file now!

src/components/Dropdown/Dropdown.tsx
```diff
@@ -39,15 +39,15 @@ export default function Dropdown(props: PropsWithChildren<{
     onToggle,
     className,
     activeClassName,
-    controlledQuery
+    controlledQuery,
```
yen-tt, line 42: nit: can we have another name for this? It seems a little unsure to me what controlledQuery would be?

src/components/SearchBar.tsx
oshi97: I kind of like how it separates responsibility here though

src/components/FilterSearch.tsx
oshi97: Ah yeah I was going to do that in a separate PR but here is fine!

src/components/Dropdown/Dropdown.tsx
oshi97: sure, this isn't exactly controlling the Dropdown's value anyways, is parentQuery ok?

yen-tt: maybe submittedQuery?

src/components/SearchBar.tsx
yen-tt: hmm okie, I'm fine with that!

src/components/Dropdown/Dropdown.tsx
oshi97: eh that's make me think this is being used for all submissions (not that parentQuery makes me think of something useful)

yen-tt: hmm i'm also not the best person at naming hahah, hard to think of a good one for this. If others don't have an opinion on this as well, we can just keep parentQuery

yen-tt (approved):
lgtm!

PR #54, opened 2022-01-27
Support custom card component

yen-tt:
This pr add the ability to add and use custom card component

- add a jambo command to create custom card component
   - command requires the new card component's name and the card template's file path.
   - will create a new component file in 'cards' directory and update card registry to include the new component
- Adjust typescript interfaces to support custom card component. Previously we intend to construct a union of all card component props when defining what a card config should be. Now that configuration interface is switch to Record<string, unknown> to support any props from custom card component.

J=SLAP-1807
TEST=manual

- in /pagebuilder folder, ran `npx jambo card --name RandomCard --cardTemplateFilePath ../components/cards/StandardCard.tsx` and see 'RandomCard.tsx' is created properly and content in 'componentRegistry.ts' is updated properly
- ran commands with wrong param values and see the right error is thrown

### DIFF
--- a/src/pagebuilder/commands/cardcreator.js
+++ b/src/pagebuilder/commands/cardcreator.js
@@ -1,0 +1,128 @@
+const fs = require('fs-extra');
+const path = require('path');
+const UserError = require('./helpers/errors/usererror');
+const { ArgumentMetadata, ArgumentType } = require('./helpers/utils/argumentmetadata');
+
+/**
+ * CardCreator represents the `card` custom jambo command.
+ * The command creates a new, custom card in the top-level 'cards' directory
+ * of a jambo repo.
+ */
+class CardCreator {
+  constructor(jamboConfig) {
+    this.config = jamboConfig;
+    this._customCardsDir = 'cards';
+  }
+
+  /**
+   * @returns {string} the alias for the create card command.
+   */
+  static getAlias() {
+    return 'card';
+  }
+
+  /**
+   * @returns {string} a short description of the create card command.
+   */
+  static getShortDescription() {
+    return 'add a new card for use in the React app';
+  }
+
+  /**
+   * @returns {Object<string, ArgumentMetadata>} description of each argument for 
+   *                                             the create card command, keyed by name
+   */
+  static args() {
+    return {
+      'name': new ArgumentMetadata(ArgumentType.STRING, 'name for the new card', true),
+      'cardTemplateFilePath': new ArgumentMetadata(ArgumentType.STRING, 'file path of card template', true)
+    };
+  }
+
+  /**
+   * @returns {Object} description of the card command
+   */
+  static describe(jamboConfig) {
+    return {
+      displayName: 'Add Card',
+      params: {
+        name: {
+          displayName: 'Card Name',
+          required: true,
+          type: 'string'
+        },
+        cardTemplateFilePath: {
+          displayName: 'Card Template\' File Path',
+          required: true,
+          type: 'string'
+        }
+      }
+    };
+  }
+
+  /**
+   * Executes the create card command with the provided arguments.
+   * 
+   * @param {Object<string, string>} args The arguments, keyed by name 
+   */
+  execute(args) {
+    this._create(args.name, args.cardTemplateFilePath);
+  }
+
+  /**
+   * Creates a new, custom card in the top-level 'Cards' directory. This card
+   * will be based off the card template from the given file path.
+   * 
+   * @param {string} newComponentName The name of the new card. A React component
+   *                                  file with this name will be created.
+   * @param {string} cardTemplateFilePath The file path of the existing card in which
+   *                                      the new one will be based on.
+   */
+  _create(newComponentName, cardTemplateFilePath) {
+    if (!(/^[A-Z][\w_]+$/.test(newComponentName))) {
+      throw new UserError(`${newComponentName} is not a valid React component name.`);
+    }
+    const newCardFilePath = `${this._customCardsDir}/${newComponentName}.tsx`;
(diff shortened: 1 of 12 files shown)


src/pagebuilder/commands/cardcreator.js
```diff
@@ -0,0 +1,128 @@
+const fs = require('fs-extra');
```
oshi97, line 1: I'm probably just missing a lot of context but is PB using jambo?

yen-tt: From a discussion with Tom, user will not be running jambo in PB. Instead, if they want to create a new card, they would have to do it through jambo in local terminal, make changes, and restart their session in PB

src/pagebuilder/templates/componentRegistry.ts
```diff
@@ -1,9 +1,13 @@
...
+  Standard = 'Standard'
+}
+
 /**
  * Map card name to the corresponding React component.
  */
-export const CardNameToComponentMap: Record<string, CardComponent> = {
-  'STANDARD': StandardCard
+export const CardRegistry: Record<string, CardComponent> = {
+  [CardTypes.Standard]: StandardCard
```
oshi97, line 12: how are cards added here?

yen-tt: jambo command will update this file to add the custom card import statement, and include custom card in the enum and in CardRegistry.

src/pagebuilder/commands/cardcreator.js
```diff
@@ -0,0 +1,128 @@
...
+    return {
+      displayName: 'Add Card',
+      params: {
+        name: {
+          displayName: 'Card Name',
+          required: true,
+          type: 'string'
+        },
+        cardTemplateFilePath: {
+          displayName: 'Card Template\' File Path',
```
nmanu1, line 55: nit: how come there's an escaped single quote in the displayName?

src/pagebuilder/commands/cardcreator.js
```diff
@@ -0,0 +1,128 @@
...
+  /**
+   * Registers new component to the card registry by directly modify content in componentRegistry.ts
+   * 
+   * @param {string} newComponentName 
+   */
+  _updateCardRegistry(newComponentName) {
+    const registryFilePath = './templates/componentRegistry.ts';
+    const registryContent = fs.readFileSync(registryFilePath).toString();
+    const newRegistryContent = `import { ${newComponentName} } from "../cards/${newComponentName}";\n`
+      + registryContent.replace(/(export enum CardTypes {\n)(.*)/gs, `$1\t${newComponentName} = '${newComponentName}',\n$2`)
```
oshi97, line 123: I think it would be nice if we didn't have to use regex here but I couldn't think of anything reasonable

src/pagebuilder/commands/cardcreator.js
```diff
@@ -0,0 +1,128 @@
...
+
+  /**
+   * Registers new component to the card registry by directly modify content in componentRegistry.ts
+   * 
+   * @param {string} newComponentName 
+   */
+  _updateCardRegistry(newComponentName) {
+    const registryFilePath = './templates/componentRegistry.ts';
+    const registryContent = fs.readFileSync(registryFilePath).toString();
+    const newRegistryContent = `import { ${newComponentName} } from "../cards/${newComponentName}";\n`
```
nmanu1, line 122: can we pass `newCardFilePath` to this function to be used here so we don't have to reconstruct the path for the import?

src/pagebuilder/commands/cardcreator.js
oshi97: Did you consider anything like having this command write to a json file instead of componentRegistry.ts directly, and then having the componentRegistry do dynamic imports. This would also offer code splitting. I was struggling thinking of how an async component would work but then remembered React.Lazy exists.

src/pagebuilder/commands/cardcreator.js
yen-tt: so the display name would be `Card Template's File Path`. Unless you think `Card Template File Path` is fine

oshi97: it looks like an "s" is missing

src/pagebuilder/commands/cardcreator.js
yen-tt: newCardFilePath contains file extension which import statement doesn't need. But I can pass it in and perform substring on it

src/pagebuilder/commands/cardcreator.js
yen-tt: ah yea

src/pagebuilder/commands/cardcreator.js
tmeyer2115: Yen and I did talk about using dynamic imports and `lazy`. We decided to go with this simpler approach for now, especially because we're not supporting hot-reloading when a card is added. FWIW, I'm open to changing this in the future if we want to. We're making a prototype for people to play with, so we can keep it simple for now and update later. 

One thing to consider here is readability. We want the resulting repo to be easily navigable and understandable by web dev who may be maintaining the site. Having an explicit registry file showing them all their card types and the mapping of type to Component may be more user friendly and modifiable than complex dynamic importing logic.

yen-tt: I attempted React.lazy and dynamic imports before this, but after talking to tom, he prefers this approach compare to the complexity involved with the first suggestion.

oshi97: that's fair I just didn't really like writing to a ts file directly, but we're not doing anything crazy here

oshi97 (approved):

PR #55, opened 2022-01-27
Fix css composition method in Facets

yen-tt:
- ensures that cssCompositionsMethod specified in Facets is also pass to individual Facet component
- update renderCheckboxOption to directly is the cssClasses props. Parent component built-in css should combine with builtin checkboxOption css to take advantage of the parent component's cssCompositionMethod before passing into renderCheckboxOption

J=none
TEST=manual

pass in customCssClasses with `label` field and cssCompositionMethod, see that Facet label styling appear as expected

### DIFF
--- a/src/components/Facets.tsx
+++ b/src/components/Facets.tsx
@@ -75,7 +75,7 @@ export default function Facets (props: FacetsProps): JSX.Element {
           <Facet
             facet={facet}
             {...config}
-            customCssclasses={cssClasses}
+            customCssClasses={cssClasses}
             cssCompositionMethod={cssCompositionMethod}
             onToggle={handleFacetOptionChange} />
           {!isLastFacet && <Divider customCssClasses={{ divider: cssClasses.divider }} cssCompositionMethod='replace'/>}
--- a/src/components/Facets.tsx
+++ b/src/components/Facets.tsx
@@ -76,6 +76,7 @@ export default function Facets (props: FacetsProps): JSX.Element {
             facet={facet}
             {...config}
             customCssclasses={cssClasses}
+            cssCompositionMethod={cssCompositionMethod}
             onToggle={handleFacetOptionChange} />
           {!isLastFacet && <Divider customCssClasses={{ divider: cssClasses.divider }} cssCompositionMethod='replace'/>}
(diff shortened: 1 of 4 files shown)


src/components/Facets.tsx
```diff
@@ -76,6 +76,7 @@ export default function Facets (props: FacetsProps): JSX.Element {
             facet={facet}
             {...config}
             customCssclasses={cssClasses}
```
nmanu1, line 78: can this also be changed to `customCssClasses` instead of `customCssclasses`?

yen-tt: what do you mean?

yen-tt: ah gotcha, updated

tmeyer2115 (approved):

src/components/StaticFilters.tsx
```diff
@@ -72,7 +72,8 @@ export default function StaticFilters(props: StaticFiltersProps): JSX.Element {
               return renderCheckboxOption({
                 option: { id: `${index}`, label: option.label },
                 onClick: selected => handleFilterOptionChange(filter, selected),
-                selected: getOptionSelectStatus(option)
+                selected: getOptionSelectStatus(option),
+                customCssClasses: cssClasses
```
nmanu1, line 76: do we also need a cssCompositionMethod here?

yen-tt: useComposedCssClasses is a hook that can only be use in other hooks and component class/functions, which renderCheckboxOption is not so we wouldn't pass cssCompositionMethod or call useComposedCssClasses.

yen-tt: actually, I updated so that the builtInCss for checkbox should be merge in at the parent component built-in level, and whatever css pass into renderCheckboxOption will be use as is. So, the cssCompositionMethod from parent component would be in effect!

nmanu1: ok! since that's abstracted out, could we change `customCssClasses` to `cssClasses` in `renderCheckboxOption`?

yen-tt: sure

PR #56, opened 2022-01-27
Setup component configuration

cea2aj:
Setup component configuration for card mappings, sorting, and general component prop setting

For the data mappings, I'm using [lodash get](https://www.geeksforgeeks.org/lodash-_-get-method/) which lets us define the data mappings with a string of fields separated by periods (e.g. 'address.city'). Each card has default data mappings but they can be customized through props (and by extension the AppConfig). Fallbacks are also supported by passing in an array ( e.g. ['address.city', 'address.country'].

For sorting, I added a concept of a setup callback to `usePageSetupEffect` so that we can set the sort after the state is cleared but before the search is ran. Currently sorting must be configured per vertical.

For the component configuration, I created a `ConfiguredSearchBar` and a `ConfiguredStandardCard`. These are HOCs which get the component config from the AppConfig, perform any necessary transformations, and render the associated component. Currently no transformations are needed since the config is just strings.

Configured components look at config for specific vertical/universal pages, and also config set in 'common'. The specific vertical/universal pages config takes precedence over the common config.

J=SLAP-1787
TEST=manual

Change the sorting config and see the ordering of the results change. Modify the card mappings in the config and see the data displayed on the card change. Add a fallback field mapping and see the fallback working when the first piece of data isn't present. Modify the search bar placeholder on a specific vertical and observe the placeholder change for just that vertical


src/components/cards/StandardCard.tsx
```diff
@@ -57,11 +61,39 @@ function isCtaData(data: unknown): data is CtaData {
  * @param props - An object containing the result itself.
  */
 export function StandardCard(props: StandardCardProps): JSX.Element {
-  const { configuration, result, customCssClasses, cssCompositionMethod } = props;
+  const { fieldMappings: customfieldMappings, showOrdinal, result, customCssClasses, cssCompositionMethod } = props;
   const cssClasses = useComposedCssClasses(builtInCssClasses, customCssClasses, cssCompositionMethod);
 
-  const cta1 = isCtaData(result.rawData.c_primaryCTA) ? result.rawData.c_primaryCTA : undefined;
-  const cta2 = isCtaData(result.rawData.c_secondaryCTA) ? result.rawData.c_secondaryCTA : undefined;
+  const defaultfieldMappings: Record<string, FieldMapping> = {
```
tmeyer2115, line 67: I think one benefit of keeping the `StandardCardConfig` is that it would allow us to define a `defaultConfig: StandardCardConfig` at the top of the file. That would include these default mappings and whatever the default is for `showOrdinal`. Organizationally,  I like having the `Props` interfaces and the defaults colocated at the top of the file. I think that improves readability. 

Open to discussion on this though.

src/components/cards/StandardCard.tsx
```diff
@@ -57,11 +61,39 @@ function isCtaData(data: unknown): data is CtaData {
...
+      apiName: 'c_secondaryCTA'
+    },
+  }
+
+  const untypedData = collectData(result.rawData, {
+    ...defaultfieldMappings,
+    ...customfieldMappings
+  });
+
+  const data = {
```
tmeyer2115, line 91: Could we make a method that handles this casting? And also maybe define a type/interface for `data`?

src/components/utils/collectData.ts
```diff
@@ -0,0 +1,74 @@
...
+  mappingType: 'CONSTANT',
+  value: string
+}
+
+export type FieldMappingAPI = {
+  mappingType: 'FIELD',
+  apiName: string | string[]
+}
+
+export type FieldMapping = FieldMappingConstant | FieldMappingAPI
```
tmeyer2115, line 13: I'm not fully sold on Rose's naming here. The terminology of `FieldMapping` makes it sound like I'm making something to a profile field.

src/components/utils/collectData.ts
```diff
@@ -0,0 +1,74 @@
...
+ * { 
+ *    question: 'Which AI search platform should we leverage?',
+ *    answer: 'Yext'
+ *  }
+ * 
+ * @param rawData The rawData from an {@link Result}
+ * @param fieldMappings Indicates where data is located within the rawData field
+ * @returns An object of fields to data
+ */
+export function collectData<FieldMappingObject extends Partial<Record<string, FieldMapping>> | undefined> (
```
tmeyer2115, line 56: I think we need a more specific name here.

src/models/cardComponent.ts
```diff
@@ -1,24 +1,19 @@
...
-export type CardConfigTypes = StandardCardConfig;
 
 /**
  * CardComponent and the corresponding config options
  */
-export interface CardConfig extends CardConfigTypes {
-  CardComponent: CardComponent
+export interface CardConfig {
+  CardComponent: CardComponent,
+  [otherProps: string]: unknown
```
tmeyer2115, line 8: Is there a name besides `otherProps` that we could use here? Something more descriptive. As it stands, this file exports two interfaces that look a heck of a lot alike. As a reader, I'm somewhat struggling to tell the difference between the two.

package.json
```diff
@@ -42,6 +42,7 @@
     "fs-extra": "^9.0.1",
     "html-webpack-plugin": "4.5.0",
     "identity-obj-proxy": "3.0.0",
+    "lodash": "^4.17.21",
```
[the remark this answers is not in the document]
cea2aj: I went with the full lodash because the lodash docs say that the use of the per method packages are discouraged and will be removed in v5. Apparently it can result in larger bundles: https://lodash.com/per-method-packages

However it looks like it's best to import 'lodash/get' and rely on treshaking, so I can update to that

src/components/cards/StandardCard.tsx
```diff
@@ -57,11 +61,39 @@ function isCtaData(data: unknown): data is CtaData {
...
+      mappingType: 'FIELD',
+      apiName: 'c_primaryCTA'
+    },
+    cta2: {
+      mappingType: 'FIELD',
+      apiName: 'c_secondaryCTA'
+    },
+  }
+
+  const untypedData = collectData(result.rawData, {
```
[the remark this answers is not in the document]
cea2aj: sure, maybe transformedRawData?

src/components/cards/StandardCard.tsx
cea2aj: Got it. I was following the change that Yen made in #54, but a `StandardCardConfig` does make sense from an organizational perspective. In my mind, the Config of a component is the props, so the reason for the separation wasn't clear, however I'm open to reverting back to improve organization

src/components/utils/collectData.ts
cea2aj: Lodash refers to it as a 'path', so maybe something along those lines such as DataPath? Or FieldDataPath

src/components/utils/collectData.ts
cea2aj: how about `processFieldData`?

src/models/cardComponent.ts
cea2aj: Sure, I can take a look at reworking this interface

package.json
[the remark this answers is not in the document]
tmeyer2115: Sgtm

src/components/cards/StandardCard.tsx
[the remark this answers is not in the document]
tmeyer2115: Sure, that works

src/components/cards/StandardCard.tsx
tmeyer2115: I'll admit, the distinction is a bit hazy. Maybe as a middle ground, we can just move this block out of the `StandardCard` and after the `StandardCardProps` declaration.

src/components/utils/collectData.ts
tmeyer2115: Yeah, I like that.

src/components/utils/collectData.ts
tmeyer2115: What about using the verb `apply`? We're applying mappings to the raw object.

src/components/cards/StandardCard.tsx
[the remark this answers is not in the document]
cea2aj: I went with transformedFieldData, but let me know if you prefer transformedRawData

src/components/cards/StandardCard.tsx
cea2aj: added validateData! The interface is defined by the provided type guards

src/models/cardComponent.ts
cea2aj: I changed this to `additionalProps` for now. I spent some time looking into reworking this interface. I think it would be good to remove `[additionalProps: string]: unknown` because it weakens the type checking. I think it would be better to pass around a CardComponent with the config set via a HOC rather than passing the config around because it would make the code simpler, and we'd be able to remove the `additionalProps`. Additional components should be able to extend the interface so long as the additional props are optional. I would like to defer that to a separate item though because it would affect many different components and I'd need to verify it works with the card creation code

src/components/utils/applyFieldMappings.ts
```diff
@@ -0,0 +1,74 @@
...
+ * { 
+ *    question: 'Which AI search platform should we leverage?',
+ *    answer: 'Yext'
+ *  }
+ * 
+ * @param rawData The rawData from an {@link Result}
+ * @param fieldMappings Indicates where data is located within the rawData field
+ * @returns An object of fields to data
+ */
+export function applyFieldMappings<FieldMappingObject extends Partial<Record<string, FieldData>> | undefined> (
```
yen-tt, line 56: Does the typing needs to be a Partial or can it just be `FieldMappingObject extends<Record<string, FieldData> | undefined`? I noticed you just type cast it below right after checking for undefined.

src/components/utils/validateData.ts
```diff
@@ -0,0 +1,41 @@
...
+  const validatedData: ValidatedData<TypeGuards> = {};
+
+  Object.entries(typeGuards).forEach(([key, typeCheck]) => {
+    if (typeCheck(data[key])) {
+      validatedData[key as keyof TypeGuards] = data[key];
+    }
+  });
+
+  return validatedData;
+}
```
yen-tt, line 41: Can we can directly use TypeGuardRecord instead of using `TypeGuards extends TypeGuardRecord`?

src/components/utils/validateData.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export function isNumber (data: any): data is number {
+  return typeof data === 'number';
+}
+
+/**
+ * Returns data which passes the corresponding typeGuard validation
+ */
+export function validateData<TypeGuards extends TypeGuardRecord>(data: any, typeGuards: TypeGuards): ValidatedData<TypeGuards> {
```
yen-tt, line 31: Can we can directly use TypeGuardRecord in this function instead of using TypeGuards which extends TypeGuardRecord?

src/components/cards/StandardCard.tsx
```diff
@@ -54,11 +82,20 @@ function isCtaData(data: unknown): data is CtaData {
...
   const cssClasses = useComposedCssClasses(builtInCssClasses, customCssClasses, cssCompositionMethod);
 
-  const cta1 = isCtaData(result.rawData.c_primaryCTA) ? result.rawData.c_primaryCTA : undefined;
-  const cta2 = isCtaData(result.rawData.c_secondaryCTA) ? result.rawData.c_secondaryCTA : undefined;
+  const transformedFieldData = applyFieldMappings(result.rawData, {
+    ...defaultFieldMappings,
+    ...customFieldMappings
+  });
+
+  const data = validateData(transformedFieldData, {
```
yen-tt, line 93: huh approach of getting typing is pretty neat

src/components/utils/validateData.ts
```diff
@@ -0,0 +1,41 @@
...
+  const validatedData: ValidatedData<TypeGuards> = {};
+
+  Object.entries(typeGuards).forEach(([key, typeCheck]) => {
+    if (typeCheck(data[key])) {
```
oshi97, line 35: what happens if the typecheck fails?

src/pagebuilder/utils/createSortBy.ts
```diff
@@ -0,0 +1,26 @@
...
+  'ASC': Direction.Ascending,
+  'DESC': Direction.Descending
+};
+
+/**
+ * Creates a SortBy from a SortOption
+ * @param sortOption 
+ * @returns SortBy
+ */
+export function createSortBy (sortOption: SortOption): SortBy {
```
yen-tt, line 20: I thinkkk SortOption and SortBy type is essentially the same. Can we just use SortBy then?

src/components/utils/applyFieldMappings.ts
cea2aj: I can actually remove the `| undefined` part. The reason for the partial is that it allows the customFieldMappings from the card which may be undefined to be accepted. We could add some complexity to the card, but I'd prefer to keep the card simple. I can remove the generic, and the type cast though

src/components/utils/validateData.ts
cea2aj: The validators may be different shapes with different keys and different expected data types. If we remove TypeGuards, we will lose the information about those specific keys and types associated with those type guards. We wouldn't be able to infer the types from that

src/components/utils/validateData.ts
cea2aj: The data associated with that key isn't set. I discussed with Tom and we're going to add a console warning.

src/pagebuilder/templates/VerticalStandardPage.tsx
```diff
@@ -18,14 +20,17 @@ interface VerticalStandardPageProps {
 
 export default function VerticalStandardPage(props: VerticalStandardPageProps) {
   const { verticalKey, cardComponent, navLinks } = props;
-  usePageSetupEffect(verticalKey);
-
   const answersAppContext = useAnswersAppContext();
+  const answersActions = useAnswersActions();
+
+  usePageSetupEffect(verticalKey, () => {
+    const sortOptions = answersAppContext.verticals?.[verticalKey ?? ''].sorting;
```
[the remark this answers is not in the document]
cea2aj: good point, i'll update to use the ternary operator

PR #58, opened 2022-01-31
add clear button to SearchBar

oshi97:
TEST=manual

clicking the searchbar clears the query and runs a search
clear button and bar icon are only visible when there is a non blank query

### DIFF
--- a/src/components/SearchBar.tsx
+++ b/src/components/SearchBar.tsx
@@ -42,7 +42,7 @@ const builtInCssClasses: SearchBarCssClasses = {
   searchButtonContainer: ' w-8 h-full mx-2 flex flex-col justify-center items-center',
   submitButton: 'h-7 w-7',
   focusedOption: 'bg-gray-100',
-  clearButton: 'mr-3.5',
+  clearButton: 'mr-3.5 cursor-pointer',
   barIcon: 'mr-0.5',
   recentSearchesOptionContainer: 'flex items-center h-6.5 px-3.5 py-1.5 cursor-pointer hover:bg-gray-100',
   recentSearchesIcon: 'w-5 mr-1 text-gray-300',
@@ -252,7 +252,8 @@ export default function SearchBar({
   function renderClearButton() {
     return (
       <>
-        <button
+        <div
+          aria-label='Clear the search bar'
           className={cssClasses.clearButton}
           onClick={() => {
             updateEntityPreviews('');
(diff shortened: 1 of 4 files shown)


src/components/SearchBar.tsx
```diff
@@ -245,7 +247,26 @@ export default function SearchBar({
...
-  const hasItems = !!(autocompleteResponse?.results.length  || (!isVertical && filteredRecentSearches?.length));
+  function renderClearButton() {
+    return (
+      <>
+        <button
+          className='mr-3.5'
+          onClick={() => {
+            updateEntityPreviews('');
+            answersActions.setQuery('');
+            executeQuery();
```
nmanu1, line 258: why is `executeQuery` called here? does clearing the text conduct a search?

src/components/SearchBar.tsx
```diff
@@ -245,7 +247,26 @@ export default function SearchBar({
     ))
   }
 
-  const hasItems = !!(autocompleteResponse?.results.length  || (!isVertical && filteredRecentSearches?.length));
+  function renderClearButton() {
+    return (
+      <>
+        <button
+          className='mr-3.5'
```
cea2aj, line 254: We probably want to include this styling and the styling below in the css class interface

src/components/SearchBar.tsx
oshi97: I asked Rose because it also seemed weird to me, she said we can just try this for now (it's what the SDK does)

src/components/SearchBar.tsx
oshi97: updated!

src/components/SearchBar.tsx
```diff
@@ -40,13 +42,13 @@ const builtInCssClasses: SearchBarCssClasses = {
   searchButtonContainer: ' w-8 h-full mx-2 flex flex-col justify-center items-center',
   submitButton: 'h-7 w-7',
   focusedOption: 'bg-gray-100',
-
+  clearButton: 'mr-3.5',
+  barIcon: 'mr-0.5',
```
cea2aj, line 46: What do you think about renaming `bar` to `verticalDivider`? It wasn't clear to me what bar was

src/components/SearchBar.tsx
```diff
@@ -245,7 +249,28 @@ export default function SearchBar({
     ))
   }
 
-  const hasItems = !!(autocompleteResponse?.results.length  || (!isVertical && filteredRecentSearches?.length));
+  function renderClearButton() {
+    return (
+      <>
+        <div
```
cea2aj, line 255: Did switching from a button to a div with the button role improve wcag?

cea2aj (approved):
🚀

PR #59, opened 2022-01-31
Minor changes for searchbar's entity preview and key nav logic

yen-tt:
- add `onFocusItem` to update entity previews when user navigate through the autocomplete options.
- when a page is loaded with initial query, update `LastTypedOrSubmittedValue` to be query to parentQuery for key nav to work

J=SLAP-1847
TEST=manual

- navigate through autocomplete options and see entity preview render
- go to a page with initial query, and see that pressing up/down arrow would update search bar and highlight the option


cea2aj (approved):
LGTM!

PR #60, opened 2022-02-01
Entity preview navigation

yen-tt:
Add entity preview navigation

expected behavior after discussion with product:
- update entity previews based on whatever user typed. Remain static when user navigate through autocomplete options.
- hitting down arrow should give you access to said entity previews, and when you reach the end you loop back to the top

Changes:
- update EntityPreviews' children in SampleVisualSearchBar to encapsulate the card with DropdownItem component so it can be navigable. Modify renderEntityPreviews to pass in a default onSubmit for when user click on entity preview item.
- update useSynchronizedRequest to check for component mount state before calling any set state.
- some cleanup for SearchBar css interface
- also fix existing wcag issues. By default, the focus element id (for aria-activedescendant) should point to input element. Add id to input element would get rid of the `aria-valid-attr-value` wcag problem

J=SLAP-1793
TEST=manual

- type arlington/covid, and see the corresponding entity preview of faq card and event card shows up.
- press arrow down through the dropdown, see that I can navigate to entity previews and loop back up at the end.
- See that clicking/entering on an entity preview execute query and bring to the corresponding vertical page.

### DIFF
--- a/src/components/SearchBar.tsx
+++ b/src/components/SearchBar.tsx
@@ -284,6 +284,13 @@ export default function SearchBar({
         activeClassName={activeClassName}
         screenReaderText={screenReaderText}
         parentQuery={query}
+        onToggle={(isActive, value = '') => {
+          if (!isActive) {
+            updateEntityPreviews(value);
+            answersActions.setQuery(value);
+            autocompletePromiseRef.current = executeAutocomplete();
+          }
+        }}
       >
         <div className={cssClasses?.inputContainer}>
           <div className={cssClasses.logoContainer}>
--- a/src/components/SearchBar.tsx
+++ b/src/components/SearchBar.tsx
@@ -38,9 +38,8 @@ const builtInCssClasses: SearchBarCssClasses = {
   inputElement: 'outline-none flex-grow border-none h-full pl-0.5 pr-2',
   logoContainer: 'w-7 mx-2.5 my-2',
   optionContainer: 'flex items-stretch py-1.5 px-3.5 cursor-pointer hover:bg-gray-100',
-  resultIconContainer: 'opacity-20 w-7 h-7 pl-1 mr-4',
   searchButtonContainer: ' w-8 h-full mx-2 flex flex-col justify-center items-center',
-  submitButton: 'h-7 w-7',
+  searchButton: 'h-7 w-7',
   focusedOption: 'bg-gray-100',
   clearButton: 'mr-3.5 cursor-pointer',
   verticalDivider: 'mr-0.5',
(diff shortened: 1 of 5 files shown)


src/components/SearchBar.tsx
```diff
@@ -282,16 +284,6 @@ export default function SearchBar({
         activeClassName={activeClassName}
         screenReaderText={screenReaderText}
         parentQuery={query}
-        onToggle={(isActive, value = '') => {
```
oshi97, line 285: the point of onToggle here was for a case like the below (I think this exact example should still work)
1. some user types in a query "c", which has the autocomplete options "covid-19" and "cawfee"
2. user clicks on "cawfee", this runs a search and also closes the dropdown
3. user clicks on the searchbar again. Briefly, the 2 autocomplete options from the original "c" query are displayed. After a short delay these change to the autocomplete options for the current query, "cawfee", which just one option that also says "cawfee"

Taking a quick look at the code I think removing the onToggle would bring this issue back? Originally, I updated autocomplete results when the dropdown was closed in order to "preload" the next autocomplete results we would want to display. If this is causing issues, I think you could also just clear out the autocomplete results when the dropdown is closed, instead of preloading the next ones? This would introduce whatever small latency an autocomplete request would take but it's probably not a huge deal and also is the current behavior of the SDK.

oshi97:
it sounds like we're just not doing left/right arrow keys?

src/components/SearchBar.tsx
yen-tt: I tried adding the onToggle back and run through the case you gave, looks like the flash from previous query to new query's autocomplete is still there? I will try clearing out results and see what happen

yen-tt:
> it sounds like we're just not doing left/right arrow keys? (edit: assuming we're not it seemed like an almost impossible task to have a general component that just "support" that behavior anyways - too many degrees of freedom. Heck google.com doesn't even do it.)

yup, we are not doing left-right arrow keys anymore. User would have to implement it if they want that behavior

src/components/SearchBar.tsx
yen-tt: Update: I added onToggle back. The unmounted issue lies in calling setSynchronizedResponse() (for autocomplete results) to trigger a render update when we already start to unmount the components and switch to new page. I end up using useComponentMountStatus(). However, the issue you describe above is still here with the onToggle. I did a little debugging and it looks like the issue is in what we expect the `value` field pass to onToggle?.(willBeOpen, value) should be. For instance, when user click into search bar, type 'c' and click on option 'cawfee'.  onToggle is getting 'c' and not the new value 'cawfee' so we are not really "preloading" the next results right now. This is outside of this item but we should fix it in another pr

nmanu1 (commented):
Now that entity previews are navigable, should they be announced by the screen reader?

src/components/SearchBar.tsx
oshi97: yeah my bad this didn't work properly, my ego wants to say this worked initially and I lost it mid PR but that's probably not the case

oshi97: actually I think I conflated the test scenario lol. this prevents the flashing when you focus an item with the arrow keys, but not if you select one with the cursor (so change step 2 above to arrow key navigation instead of a click). I tested this real quick on current develop and think this is still the case? we still probably want to handle this for clicking an option without first focusing it though, another item sgtm!

oshi97 (approved):

yen-tt:
> Now that entity previews are navigable, should they be announced by the screen reader?

Yup. But counting the entity previews would be a little more work compare to adding the recent search announcement previously. We should address this in another item/pr

PR #61, opened 2022-02-02
Support SCSS customization

nmanu1:
Adds support for customizing the app's styling by changing the common styles in the PB config. The style variables are computed in the `PageBuilderApp` any time the config changes. Refreshing the page on the app displays the updated styling.

For now, the styling only affects the `LocationBias`, `StandardCard`, and `SpellCheck` components. The styling for these has been translated from tailwind to SCSS because PB 2.0 will not be using tailwind.

J=SLAP-1788
TEST=manual

Check that updating a style in the config is reflected when refreshing the page.


src/pagebuilder/scripts/generateAnswersVariables.js
```diff
@@ -0,0 +1,13 @@
+const fs = require('fs');
```
yen-tt, line 1: I think this approach of watching the config file would be ok for the purpose of testing this prototype repo but it may not work with page builder 2.0 system? We have a pageBuilderAnswersAppConfig file to mimic what PB2.0 would pass to PageBuilderApp component but I assume there wouldn't be an actual json file to monitor if P.B would construct whatever json data from user interaction with their UI and pass it to PageBuilderApp component. If there's a file, would P.B be the one to invoke this script and provide the path to the config file?

cea2aj: Although the details of PageBuilder are not fully worked out yet, I think it's very likely there will be a json file. From what i've heard, page builder will produce a json file, and our job is to read that file and create an experience based on it. I find it unlikely that page builder would be passing the configuration directly to our component. If we are operating based on a json file, the systems won't be so tightly coupled

cea2aj: Also page builder 2.0 will use the repo that it operates on as the primary storage mechanism. Currently there is no plan to store the config in a database.

src/pagebuilder/scripts/createAnswersVariables.js
```diff
@@ -0,0 +1,49 @@
+const fs = require("fs-extra");
+
+const styleVariableNames = {
+  '--yxt-color-brand-primary': 'brandColor',
+  '--yxt-base-font-size': 'baseFontSize',
+  '--yxt-cta-border-radius': ['ctas', 'borderRadius'],
+  '--yxt-cta-color': ['ctas', 'color']
+};
+const path = __dirname + '/../sass/answers-variables.scss';
```
oshi97, line 9: is it necessary to stick these onto an scss file? couldn't we just import them as inline styles that we attach onto App.tsx?

src/pagebuilder/sass/StandardCard.module.scss
```diff
@@ -0,0 +1,74 @@
...
+.StandardCard {
+  &__container {
+    display: flex;
+    flex-direction: column;
+    justify-content: space-between;
+    border-width: 1px;
+    border-radius: 0.5rem;
+    margin-bottom: 1rem;
+    padding: 1rem;
+    --tw-shadow: 0 1px 2px 0 rgba(0, 0, 0, 0.05);
```
cea2aj, line 13: I don't think we can use `--tw-shadow` because it's provided by tailwind

src/pagebuilder/sass/answers-variables.scss
```diff
@@ -0,0 +1,6 @@
+:root {
```
cea2aj, line 1: nit: can we make the spacing of this file and the default file consistent?

src/pagebuilder/sass/LocationBias.module.scss
```diff
@@ -0,0 +1,26 @@
+@import './answers-variables-default';
+@import './answers-variables';
+
+.LocationBias {
```
cea2aj, line 4: If you remove the .Location bias class and instead have the other classes at the top of the file, are you able to import the styles object and pass it to the `customCssClasses` prop directly in the configured objects without mapping all of the props? If so, the naming will need to match up and I believe spellcheck_loading would need to be updated

src/pagebuilder/scripts/createAnswersVariables.js
```diff
@@ -0,0 +1,49 @@
...
+
+function createAnswersVariables(config) {
+  let answersVars = ':root {\n';
+  const styles = config.common?.style;
+  if (styles) {
+    Object.entries(styleVariableNames).forEach(([name, path]) => {
+      let value = '';
+      if (!Array.isArray(path)) {
+        value = _findValue(styles, path);
```
cea2aj, line 18: does `styles[path]` work instead of `_findValue`?

src/pagebuilder/scripts/createAnswersVariables.js
```diff
@@ -0,0 +1,49 @@
...
+
+const styleVariableNames = {
+  '--yxt-color-brand-primary': 'brandColor',
+  '--yxt-base-font-size': 'baseFontSize',
+  '--yxt-cta-border-radius': ['ctas', 'borderRadius'],
+  '--yxt-cta-color': ['ctas', 'color']
+};
+const path = __dirname + '/../sass/answers-variables.scss';
+
+function createAnswersVariables(config) {
```
cea2aj, line 11: nit: can we indicate that this creates a file with a name such as `createAnswersVariablesFile`?

src/pagebuilder/scripts/createAnswersVariables.js
```diff
@@ -0,0 +1,49 @@
...
+  let answersVars = ':root {\n';
```
cea2aj, line 12: nit: what do you think of a name such as `answersVarsFileBody`?

src/pagebuilder/scripts/createAnswersVariables.js
```diff
@@ -0,0 +1,49 @@
...
+  if (styles) {
+    Object.entries(styleVariableNames).forEach(([name, path]) => {
+      let value = '';
+      if (!Array.isArray(path)) {
+        value = _findValue(styles, path);
+      } else {
+        value = path.reduce((obj, key) => {
+          return _findValue(obj, key);
+        }, styles).toString();
+      }
```
cea2aj, line 23: nit: If we import lodash get with `import get from 'lodash/get';`, I think we can simplify this to `const value = get(styles, path);`

nmanu1: updated!

src/pagebuilder/scripts/createAnswersVariables.js
nmanu1: updated to use lodash/get instead

src/pagebuilder/sass/StandardCard.module.scss
nmanu1: updated!

src/pagebuilder/sass/LocationBias.module.scss
nmanu1: That worked. Updated!

src/pagebuilder/scripts/createAnswersVariables.js
nmanu1: the styling is being added as CSS modules in the higher order configured components. so each component's SCSS file needs to get these style variables before being imported into the configured component. adding them inline through App.tsx would only work if we could use tailwind for PB 2.0

package.json
```diff
@@ -78,7 +79,7 @@
     "workbox-webpack-plugin": "5.1.4"
   },
   "scripts": {
-    "start": "node scripts/start.js",
+    "start": "node scripts/start.js & node src/pagebuilder/scripts/generateAnswersVariables.js",
     "build": "node scripts/build.js",
```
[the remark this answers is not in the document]
nmanu1: Updated the script to accept a `-watch` flag. Without it, the watcher won't run, so it can be used for the build

src/pagebuilder/scripts/createAnswersVariables.js
oshi97: I thought that since these are css variables and not scss you can specify them anywhere and they don't need to be imported? I know this is true for regular css variables but I'm not sure if css modules do any special isolation
like you can specify something like the below anywhere in your html
`:root {
  --base-spacing: 12px;
}`
and then you can use `var(--base-spacing)` anywhere in your css or scss without importing anything

nmanu1: Oh, something like using `document.documentElement.style.setProperty()` in `PageBuilderApp` to set the variables instead?

oshi97: I didn't think of that but that could work! I just meant putting it on a component as an inline style
something like this though I am having some trouble with their steps
https://stackoverflow.com/questions/52005083/how-to-define-css-variables-in-style-attribute-in-react-and-typescript

cea2aj: Oliver, are you saying add support to all of the components so that they support inline styles? Although your idea of adding the css vars to the page and using those vars makes sense so long as css modules support that

oshi97: Ok I think something like the below will work! I don't think we even need to use setProperty
image

oshi97: I tried using the variables in css for descendant html nodes and it seems to work!

cea2aj: I see, so we read in the config, create this `cssProperties` object, set it as an inline style on a parent react component, and then all descendent html nodes should be able to use those values in their scss

oshi97: I think we might be able to get rid of the watcher this way?

nmanu1: yeah I was just trying something similar out and it works!

package.json
```diff
@@ -79,8 +79,8 @@
     "workbox-webpack-plugin": "5.1.4"
   },
   "scripts": {
-    "start": "node scripts/start.js & node src/pagebuilder/scripts/generateAnswersVariables.js",
-    "build": "node scripts/build.js",
+    "start": "node scripts/start.js & node src/pagebuilder/scripts/generateAnswersVariables.js -watch",
+    "build": "node src/pagebuilder/scripts/generateAnswersVariables.js & node scripts/build.js",
```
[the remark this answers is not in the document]
nmanu1: Removed the script after switching to inline styling

src/pagebuilder/scripts/createAnswersVariables.js
nmanu1: updated with the inline styling!

src/pagebuilder/PageBuilderApp.tsx
```diff
@@ -69,3 +74,13 @@ function constructPageRoutes(universal: UniversalPageConfig, verticals: Vertical
...
 }
+
+function createStyleVariables(config: StyleConfig): React.CSSProperties {
+  const styles: React.CSSProperties = {};
+  styleVariables.forEach(style => {
+    const value = get(config, style.path) || style.default;
+    const thisStyle = { [style.name]: value } as React.CSSProperties;
+    Object.assign(styles, thisStyle);
+  });
+  return styles;
```
cea2aj, line 85: nit: can we do this with a `reduce`?

src/pagebuilder/templates/styleVariables.ts
```diff
@@ -0,0 +1,28 @@
...
+    default: '#2563EB'
+  },
+  {
+    name: '--yxt-base-font-size',
+    path: 'baseFontSize',
+    default: '16px'
+  },
+  {
+    name: '--yxt-cta-border-radius',
+    path: ['ctas', 'borderRadius'],
```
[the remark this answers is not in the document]
nmanu1: sure, I can change it to that

src/pagebuilder/templates/styleVariables.ts
```diff
@@ -0,0 +1,28 @@
+interface StyleVariable {
+  name: string,
```
oshi97, line 2: we can enforce that the name "looks like" a css variable by changing this to
```name: `--${string}\` ```
I only learned this today from https://stackoverflow.com/questions/52005083/how-to-define-css-variables-in-style-attribute-in-react-and-typescript !

Comment (oshi97) [reply src/pagebuilder/templates/styleVariables.ts:2]: The official name is apparently Template Literal Types https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html

File: src/pagebuilder/components/ConfiguredLocationBias.tsx [hunk 15]
```diff
@@ -0,0 +1,11 @@
+import { default as LocationBias } from '../../components/LocationBias';
+import locationBiasStyles from '../sass/LocationBias.module.scss';
+
+export default function ConfiguredLocationBias() {
```
oshi97, line 4: what does Configured mean here? wondering if there's a more descriptive name

src/pagebuilder/index.tsx
```diff
@@ -13,7 +14,7 @@ import registerServiceWorker from '../registerServiceWorker';
  */
 ReactDOM.render(
   <React.StrictMode>
-    <PageBuilderApp config={pageBuilderAnswersAppConfig}/>
+    <PageBuilderApp config={pageBuilderAnswersAppConfig as AnswersAppConfig}/>
```
oshi97, line 17: is this type assertion necessary? I would have hoped typescript can figure out the static shape of the json
I found this post https://stackoverflow.com/questions/49996456/importing-json-file-in-typescript
but it looks like we already have `resolveJsonModule` set to true

src/pagebuilder/components/ConfiguredLocationBias.tsx
cea2aj: `Configured` indicates that the component is a HOC which is configured based on the page builder app config

src/pagebuilder/index.tsx
nmanu1: yeah, without the assertion, it's not able to type the json properly for things like "Standard" being an acceptable card type instead of just an arbitrary string

src/pagebuilder/templates/styleVariables.ts
nmanu1: cool! updated

src/pagebuilder/components/ConfiguredLocationBias.tsx
oshi97: gotcha ok I thought this was just setting styling but I forgot we're also adding props and stuff

oshi97 (approved):

cea2aj (approved):
🔥

PR #62, opened 2022-02-02
Serve on Yext Sites

cea2aj:
Serve on yext sites

This adds default configuration to support yext sites. It works for both full production builds and for live preview. Currently the page is being served by the dev branch, but it will be available for develop and main once it's merged into those branches. I opted for `npm install` for the `installDepsCmd` so that the package.json file can be modified in live preview and the package-lock.json can be updated as necessary. For the `buildSetupCmd`, I went with `npm ci` because it's faster and ensures we get a clean build based purely on the package-lock.json file.

J=SLAP-1841
TEST=manual

Add the repo to yext sites and view the page being served at https://devyextsites-react--site--search--starter-slapshot-pagescdn-com.preview.pagescdn.com/. Also load the experience in the code editor and see the page work with live preview. Modify a file and see the file reflected in the page after refreshing the page

### DIFF
--- a/sites-config/ci_config.json
+++ b/sites-config/ci_config.json
@@ -9,11 +9,11 @@
   },
   "baseImageTag": "node-17",
   "dependencies": {
-      "installDepsCmd": "echo npm install",
+      "installDepsCmd": "npm ci",
       "requiredFiles": []
   },
   "buildArtifacts": {
-      "buildSetupCmd": "npm install",
+      "buildSetupCmd": "npm ci",
       "buildCmd": "npm run build"
   },
   "livePreview": {
--- a/sites-config/ci_config.json
+++ b/sites-config/ci_config.json
@@ -9,11 +9,11 @@
   },
   "baseImageTag": "node-17",
   "dependencies": {
-      "installDepsCmd": "npm install",
-      "requiredFiles": []
+      "installDepsCmd": "npm ci",
+      "requiredFiles": ["package.json"]
   },
   "buildArtifacts": {
-      "buildSetupCmd": "npm install",
+      "buildSetupCmd": "npm ci",
       "buildCmd": "npm run build"
   },
   "livePreview": {
--- a/sites-config/ci_config.json
+++ b/sites-config/ci_config.json
@@ -13,7 +13,7 @@
       "requiredFiles": []
   },
   "buildArtifacts": {
-      "build_setup_cmd": "echo build setup cmd",
+      "build_setup_cmd": "npm install",
       "build_cmd": "npm run build"
   },
   "livePreview": {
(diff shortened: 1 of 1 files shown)


sites-config/ci_config.json
```diff
@@ -0,0 +1,24 @@
+{
+  "artifactStructure": {
+      "assets": [
+          {
+              "root": "build"
+          }
+      ],
+      "hbsTemplates": []
+  },
+  "base_image_tag": "node-14",
```
oshi97, line 10: can we use node 16? I think I tried 17 locally a month-ish and it didn't work, though if it did that might be even better?

sites-config/ci_config.json
```diff
@@ -0,0 +1,24 @@
...
+  "dependencies": {
+      "installDepsCmd": "npm install",
+      "requiredFiles": []
+  },
+  "buildArtifacts": {
+      "build_setup_cmd": "npm install",
```
oshi97, line 16: is installDepsCmd not run before buildArtifacts?

cea2aj: ah, you're right that we should be able to rely on the install from the `installDepsCmd`. I'll try that

sites-config/ci_config.json
cea2aj: I can give it a shot

sites-config/ci_config.json
cea2aj: turns out it supports camel case for everything!

oshi97: oh wow they support both that's pretty crazy O.o

cea2aj: When I run the install just in the `build_setup_cmd`, live preview breaks, and if I run the install just in the `install_deps_cmd`, the build breaks, so it looks like we need it for both which is what Theme 1.0 sites do

sites-config/ci_config.json
cea2aj: node 17 is working!

sites-config/ci_config.json
```diff
@@ -0,0 +1,22 @@
...
+  "dependencies": {
+      "installDepsCmd": "npm install",
+      "requiredFiles": []
+  },
+  "buildArtifacts": {
+      "buildSetupCmd": "npm ci",
+      "buildCmd": "npm run build"
+  },
+  "livePreview": {
+      "serveCmd": "export PORT=8080 && npm run start"
```
oshi97, line 20: this a super nit but I think you can just do `PORT=8080 npm start` (or npm run start, just dont think we want the export PORT)

sites-config/ci_config.json
```diff
@@ -0,0 +1,22 @@
+{
+  "artifactStructure": {
+      "assets": [
+          {
```
oshi97, line 4: ahh another nit (and I should have noticed this earlier) but I think the spacing is inconsistent? think the vscode `> Format Document` should fix stuff

oshi97 (approved):

PR #63, opened 2022-02-02
fix SearchBar autocomplete options flickering for certain cases

oshi97:
updates SearchBar to clear the autocomplete whenever the dropdown closes, in order to remove stale autocomplete response data.

J=SLAP-1859
TEST=manual

autocomplete options no longer flicker
steps:
1. type "c" into search bar
2. either click on the "cawfee" autocomplete option, or submit it with the arrow keys + enter
3. click the search bar again, displays autocomplete options for "cawfee" and not "c" with no flickering
previously, would first briefly display the options for "c" before quickly changing to the options for "cawfee"

see that vertical links don't flicker, using the "locations" query and then selecting "locations that take mastercard and offer free wifi" `in Locations`

### DIFF
--- a/src/components/SearchBar.tsx
+++ b/src/components/SearchBar.tsx
@@ -144,13 +144,24 @@ export default function SearchBar({
     executeQueryWithNearMeHandling();
   }
 
+  function updateDropdownOptionsAndQuery(value: string) {
+    answersActions.setQuery(value);
+    if (renderEntityPreviews) {
+      const restrictVerticals = calculateRestrictVerticals(entityPreviews);
+      const universalLimit = calculateUniversalLimit(entityPreviews);
+      executeEntityPreviewsQuery(value, universalLimit, restrictVerticals);
+    }
+    autocompletePromiseRef.current = executeAutocomplete();
+  }
+
   const handleSubmit = (value: string, _index: number, itemData?: FocusedItemData) => {
-    answersActions.setQuery(value || '');
     if (itemData && typeof itemData.verticalLink === 'string') {
+      answersActions.setQuery(value);
       browserHistory.push(itemData.verticalLink, {
         querySource: QuerySource.Autocomplete
       });
     } else {
+      updateDropdownOptionsAndQuery(value);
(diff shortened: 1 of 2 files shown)


src/components/SearchBar.tsx
```diff
@@ -144,45 +144,40 @@ export default function SearchBar({
...
+
   const handleSubmit = (value: string, _index: number, itemData?: FocusedItemData) => {
-    answersActions.setQuery(value || '');
     if (itemData && typeof itemData.verticalLink === 'string') {
+      answersActions.setQuery(value);
       browserHistory.push(itemData.verticalLink, {
         querySource: QuerySource.Autocomplete
       });
     } else {
+      updateDropdownOptionsAndQuery(value);
```
tmeyer2115, line 164: For my own edification, when we submit a query (and don't redirect to a vertical page), we first render the entity previews (using the new search term) and fire off an auto-complete request. I'm assuming this saves us from doing that work when the `Dropdown` is actually toggled again with the new search term? What if the user doesn't toggle the `Dropdown` after submitting the new search term? In that case, will we have made unnecessary API calls (for the auto-complete data and the entity previews)?

oshi97: Yup, doing the autocomplete request ahead of time removes any latency for the next autocomplete, but it does always result in an extra request. I'll change it to clear the autocomplete data after submission instead? 

I wasn't sure which was better between the two, but initially leaned towards making extra api requests for a slightly snappier UX since our service worker caching would handle duplicate requests. I did forget that we only have service worker caching for autocomplete requests made for visual autocomplete (though we could change this) and that people also might use our components without using our service worker.

oshi97: I still think if I was a 3rd party dev making a component for yanswers that I would just make the occasional extra request whenever somebody submits or clicks away from the searchbar, especially since we don't even charge for autocomplete, but considering current circumstances I don't mind being a bit safer either.

tmeyer2115 (approved):

PR #64, opened 2022-02-02
upgrade to create-react-app 5, "un-eject"

oshi97:
This PR updates react-scripts to 5, and re-inserts (un-ejects?) the app back into the create-react-app ecosystem.
react-router-dom was not updated from 5 to 6, due to breaking changes

appliedFilters.test.tsx was moved into the `src` folder, since create-react-app only supports having test files inside the `src` folder. This preferred folder structure for tests in create-react-app is having them adjacent to the source file
they are testing. This lets tests be more discoverable and makes import paths shorter.
https://stackoverflow.com/questions/62442797/jest-folder-structure.

create-react-app also does not support the `setupFilesAfterEnv` jest config without ejecting, so `tests/setup/setup-env.ts` was moved to `src/setupTests.ts`

The third party notices github action was updated to use a macos runner instead of an ubuntu runner. This is
because fsevents is an `optionalDepedency` that is only installed when the user is on macos. Generally speaking optional dependencies are rare and only used by libraries like `chokidar` which need to do os specific things.

J=SLAP-1798
TEST=manual

npm test script works
npm test:watch script works
page loads, vertical links work, visual autocomplete works
npm build script works
service worker works with a production build
no service worker exists when using the dev flow


tailwind.config.js
```diff
@@ -1,6 +1,5 @@
 module.exports = {
-  purge: ['./src/**/*.{js,jsx,ts,tsx}', './public/index.html'],
-  darkMode: false, // or 'media' or 'class'
```
oshi97, line 3: tailwind was saying to either change `darkMode` to `media` or remove it

cea2aj (approved):
And they said couldn't be done (un-ejecting from CRA)

PR #65, opened 2022-02-02
Update screenreader announcements

nmanu1:
Update ScreenReader announcements to include entity previews now that they are navigable by arrow key. Fix a bug where entity previews weren't displayed if there were no recent searches or query suggestions.

J=SLAP-1858
TEST=manual

Check that entity previews are counted in the announcement when present.

### DIFF
--- a/src/components/SearchBar.tsx
+++ b/src/components/SearchBar.tsx
@@ -271,8 +271,9 @@ export default function SearchBar({
     );
   }
 
-  const hasItems = !!(autocompleteResponse?.results.length || (!isVertical && filteredRecentSearches?.length));
-  const screenReaderText = getScreenReaderText(autocompleteResponse?.results.length, filteredRecentSearches?.length)
+  const entityPreviewsCount = calculateEntityPreviewsCount(verticalResultsArray);
+  const hasItems = !!(autocompleteResponse?.results.length || (!isVertical && (filteredRecentSearches?.length || entityPreviewsCount)));
+  const screenReaderText = getScreenReaderText(autocompleteResponse?.results.length, filteredRecentSearches?.length, entityPreviewsCount);
   const activeClassName = classNames(cssClasses.inputDropdownContainer, {
     [cssClasses.inputDropdownContainer___active ?? '']: hasItems
   });
@@ -333,7 +334,15 @@ function StyledDropdownMenu({ cssClasses, children }: PropsWithChildren<{
   )
 }
 
-function getScreenReaderText(autocompleteOptions = 0, recentSearchesOptions = 0) {
+function calculateEntityPreviewsCount(verticalResultsArray: VerticalResults[]) {
+  let numEntityPreviews = 0;
+  verticalResultsArray.forEach((verticalResults) => {
+    numEntityPreviews += verticalResults.results.length;
+  })
+  return numEntityPreviews;
+}
+
+function getScreenReaderText(autocompleteOptions = 0, recentSearchesOptions = 0, entityPreviewsCount = 0): string {
   const recentSearchesText = recentSearchesOptions > 0
     ? processTranslation({
       phrase: `${recentSearchesOptions} recent search found.`,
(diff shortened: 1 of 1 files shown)


src/components/SearchBar.tsx
```diff
@@ -333,20 +334,38 @@ function StyledDropdownMenu({ cssClasses, children }: PropsWithChildren<{
...
-  });
-  return (recentSearchesText + ' ' + autocompleteText).trim();
+  const entityPreviewsText = entityPreviewsCount > 0
+    ? ' ' + processTranslation({
+      phrase: `${entityPreviewsCount} entity preview found.`,
+      pluralForm: `${entityPreviewsCount} entity previews found.`,
+      count: entityPreviewsCount
+    })
+    : '';
+  const autocompleteText = (autocompleteOptions > 0 || !(recentSearchesText || entityPreviewsText))
```
oshi97, line 360: I think it might be a little clearer if the case when all 3 option lengths are 0 is just put at the top, took me a second to read what this is doing

src/components/SearchBar.tsx
```diff
@@ -333,20 +334,38 @@ function StyledDropdownMenu({ cssClasses, children }: PropsWithChildren<{
   )
 }
 
-function getScreenReaderText(autocompleteOptions = 0, recentSearchesOptions = 0) {
+function calculateEntityPreviewsCount(verticalResultsArray: VerticalResults[]) {
+  let numEntityPreviews = 0;
+  verticalResultsArray.forEach((verticalResults) => {
```
oshi97, line 339: This is not necessarily the number of entity previews, since the user doesn't have to create a preview for every single result. I think we'd want to use `recursivelyMapChildren` (or even make a version that just loop through) props.children and counts the number of `DropdownItem` instances. We could probably stuff this function into `EntityPreviews.tsx` (not sure if this is the best place but the other functions live there right now)

src/components/SearchBar.tsx
```diff
@@ -271,8 +271,9 @@ export default function SearchBar({
     );
   }
 
-  const hasItems = !!(autocompleteResponse?.results.length || (!isVertical && filteredRecentSearches?.length));
-  const screenReaderText = getScreenReaderText(autocompleteResponse?.results.length, filteredRecentSearches?.length)
+  const entityPreviewsCount = calculateEntityPreviewsCount(verticalResultsArray);
+  const hasItems = !!(autocompleteResponse?.results.length || (!isVertical && (filteredRecentSearches?.length || entityPreviewsCount)));
```
oshi97, line 275: is `entityPreviewsCount` ever non-zero when the `isVertical` is true?

src/components/SearchBar.tsx
```diff
@@ -271,8 +271,9 @@ export default function SearchBar({
     );
   }
 
-  const hasItems = !!(autocompleteResponse?.results.length || (!isVertical && filteredRecentSearches?.length));
-  const screenReaderText = getScreenReaderText(autocompleteResponse?.results.length, filteredRecentSearches?.length)
+  const entityPreviewsCount = calculateEntityPreviewsCount(verticalResultsArray);
```
yen-tt, line 274: I was thinking we may want to count the number of entityPreviews (specifically what's navigable so the number of DropdownItems in it) from the results after calling transformEntityPreviews()? Since user pass in renderEntityPreviews, it give them freedom to decide what to or to not render, and if they would be navigable (if wrap in DropdownItem) so counting it after transformEntityPreviews which cover all cases with the flexibility we offer from renderEntityPreviews. But if we just want the default behavior to be assuming display all entity previews, then this would be sufficient. If we want to keep this, I think you should check if renderEntityPreviews is pass before including this count in the screen reader announcement

src/components/SearchBar.tsx
```diff
@@ -271,8 +271,9 @@ export default function SearchBar({
...
+  const hasItems = !!(autocompleteResponse?.results.length || (!isVertical && (filteredRecentSearches?.length || entityPreviewsCount)));
+  const screenReaderText = getScreenReaderText(autocompleteResponse?.results.length, filteredRecentSearches?.length, entityPreviewsCount);
```
yen-tt, line 276: I think you should check if renderEntityPreviews is pass before including this count in the screen reader announcement. Otherwise it may say there's entity previews even if user didn't include a render function to display them

src/components/SearchBar.tsx
```diff
@@ -333,20 +334,38 @@ function StyledDropdownMenu({ cssClasses, children }: PropsWithChildren<{
...
-  const autocompleteText = processTranslation({
-    phrase: `${autocompleteOptions} autocomplete suggestion found.`,
-    pluralForm: `${autocompleteOptions} autocomplete suggestions found.`,
-    count: autocompleteOptions
-  });
-  return (recentSearchesText + ' ' + autocompleteText).trim();
+  const entityPreviewsText = entityPreviewsCount > 0
+    ? ' ' + processTranslation({
+      phrase: `${entityPreviewsCount} entity preview found.`,
+      pluralForm: `${entityPreviewsCount} entity previews found.`,
```
yen-tt, line 356: what do you think about changing from `entity preview` to something like `result preview`? the term entity is more of a backend/data term I think

src/components/SearchBar.tsx
nmanu1: Updated it to use `recursivelyMapChildren` and check for `DropdownItem` instances

src/components/SearchBar.tsx
nmanu1: Updated it to get the count after `transformEntityPreviews`

src/components/SearchBar.tsx
nmanu1: `entityPreviews` already checks if there is a render function before populating this. I also updated to just count after getting the transformed entity previews

src/components/SearchBar.tsx
nmanu1: if a `renderEntitiesPreview` function were provided then I think it could be true. but in that case I think we'd probably want to include them in the count. so I'll move it out of this check

PR #66, opened 2022-02-03
fix navigation styling from tailwind 2->3 upgrade

oshi97:
Upgrading tailwind from 2 to 3 introduced a number of styling regressions
due to a number of arguably breaking but undocumented changes.

It looks like tailwind has changed how it behaves when two tailwind classes modify
the same css attribute, for example having text-blue-600 and text-gray-800 on the same element.
The Navigation component was updated to no longer rely on this kind of behavior.

I did a quick audit to check that our other components don't rely on this behavior.

Tailwind also no longer supports `border-opacity` class, [you have to specify it as part of the border color](https://github.com/tailwindlabs/tailwindcss/discussions/7320#discussioncomment-2112053), despite not being mentioned as removed in the upgrade guide.

J=SLAP-1864
TEST=manual

test side by side with local site and a version of the react-site-search-starter
hosted on yext sites that doesn't have this styling regression

the navigation's menu button (which appears when the navigation tabs no longer fully
fit within the page's width) has correct styling for it's 4 different states

navigation tab styling outside of the menu have correct styling for both active, inactive, and hover
states

### DIFF
--- a/src/components/Navigation.tsx
+++ b/src/components/Navigation.tsx
@@ -11,32 +11,40 @@ interface NavigationCssClasses {
   menuWrapper?: string,
   navLinkContainer?: string,
   navLinkContainer___active?: string,
+  navLinkContainer___inActive?: string,
   navLink?: string,
   kebabIcon?: string,
   menuButton?: string,
   menuButtonContainer?: string,
-  menuButton___menuOpen?: string,
+  menuButton___menuOpenNoActiveLink?: string,
+  menuButton___menuOpenHasActiveLink?: string,
   menuButton___hasActiveLink?: string,
+  menuButton___noActiveLink?: string,
   menuContainer?: string,
   menuNavLink?: string,
   menuNavLinkContainer?: string,
-  menuNavLinkContainer___active?: string
+  menuNavLinkContainer___active?: string,
+  menuNavLinkContainer___inActive?: string
 }
 
 const builtInCssClasses: NavigationCssClasses = {
   nav: 'border-b border-gray-200 text-gray-600 flex space-x-6 font-medium mb-6',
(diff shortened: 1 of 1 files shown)


cea2aj:
| a number of styling regressions due to a number of arguably breaking but undocumented changes

src/components/Navigation.tsx
```diff
@@ -11,32 +11,40 @@ interface NavigationCssClasses {
...
+  navLinkContainer___inActive?: string,
   navLink?: string,
   kebabIcon?: string,
   menuButton?: string,
   menuButtonContainer?: string,
-  menuButton___menuOpen?: string,
+  menuButton___menuOpenNoActiveLink?: string,
+  menuButton___menuOpenHasActiveLink?: string,
   menuButton___hasActiveLink?: string,
+  menuButton___noActiveLink?: string,
```
cea2aj, line 22: I'm just wondering if we should go with menuClosedHasActiveLink and menuClosedNoActiveLink for consistency here. Otherwise LGTM

nmanu1 (commented):
Looks like the WCAG tests are failing. Did the color contrast for some things get changed?

oshi97:
> Looks like the WCAG tests are failing. Did the color contrast for some things get changed?



> Looks like the WCAG tests are failing. Did the color contrast for some things get changed?

did a side by side of a branch with wcag passing and this branch and the color for an active nav link has the same rgb value so I'm a little confused... I'll keep looking!

src/components/Navigation.tsx
oshi97: yeah that's more consistent!

tmeyer2115 (approved):

PR #67, opened 2022-02-04
Fix VA styling

nmanu1:
Fix a styling regression where the magnifying glass icon for query suggestions in VA no longer appeared.

Also, fix the styling for entity previews when no other items are in the dropdown. In this case, the divider should not appear for the entity previews of the first vertical. But, if there are other verticals with entity previews, then these should have dividers.

J=SLAP-1861
TEST=manual

See that the icons for VA are correct and the divider above entity previews only appears over a vertical's preview cards if there are other dropdown items before it.

### DIFF
--- a/src/components/EntityPreviews.tsx
+++ b/src/components/EntityPreviews.tsx
@@ -5,7 +5,7 @@ import recursivelyMapChildren from './utils/recursivelyMapChildren';
 
 interface EntityPreviewsProps {
   verticalKey: string,
-  children: (results: Result[], showDivider?: boolean) => JSX.Element,
+  children: (results: Result[], index: number) => JSX.Element,
   limit?: number
 }
 
@@ -24,9 +24,9 @@ export default function EntityPreviews(_: EntityPreviewsProps) {
 /**
  * Recursively passes vertical results into instances of EntityPreview.
  */
-export function transformEntityPreviews(entityPreviews: JSX.Element, verticalResultsArray: VerticalResults[], showDivider?: boolean): ReactNode {
+export function transformEntityPreviews(entityPreviews: JSX.Element, verticalResultsArray: VerticalResults[]): ReactNode {
   const verticalKeyToResults = getVerticalKeyToResults(verticalResultsArray);
-  let hasEntityPreviews = false;
+  let index = 0;
   const renderedChildren = recursivelyMapChildren(entityPreviews, child => {
     if (!isValidElement(child) || child.type !== EntityPreviews) {
       return child;
@@ -35,10 +35,7 @@ export function transformEntityPreviews(entityPreviews: JSX.Element, verticalRes
     if (!(verticalKey in verticalKeyToResults)) {
       return null;
     }
-
-    const shouldShowDivider = hasEntityPreviews || showDivider;
-    hasEntityPreviews = true;
-    return children(verticalKeyToResults[verticalKey], shouldShowDivider);
+    return children(verticalKeyToResults[verticalKey], index++);
   });
   return renderedChildren;
 }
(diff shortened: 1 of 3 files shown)


src/components/EntityPreviews.tsx
```diff
@@ -5,7 +5,7 @@ import recursivelyMapChildren from './utils/recursivelyMapChildren';
 
 interface EntityPreviewsProps {
   verticalKey: string,
-  children: (results: Result[]) => JSX.Element,
+  children: (results: Result[], showDivider?: boolean) => JSX.Element,
```
oshi97, line 8: showDivider is a little too specific to what we happen to be doing in `SampleVisualSearchBar`, is there a more  generic way to pass this information down?

nmanu1: to add dividers between verticals, the children need some way of knowing if there are other verticals with entity previews being rendered before them (in which case they should add a divider). so instead, I could pass an index here and then in `renderEntityPreviews` each child would only add a divider if it's not the first child.

but then, to add the initial divider between the first part of the dropdown and the entity previews, the divider would either need to be added in the `SearchBar` component, or `showFirstDivider` could be passed to `renderEntityPreviews`. is there a preference for where this divider should live (in `SearchBar` or `SampleVisualSearchBar`)?

src/components/SampleVisualSearchBar.tsx
```diff
@@ -84,7 +86,6 @@ function FaqCard({ result }: CardProps) {
     : [];
 
   return (<div key={faqData.question}>
-    <div className='h-px bg-gray-200 mt-1 mb-4 mx-2.5'></div>
     <div tabIndex={0} className='flex flex-row mx-4 mb-3 rounded-md'>
```
cea2aj, line 89: Do we not need a divider between each FAQ?

src/components/EntityPreviews.tsx
oshi97: Passing an index sounds good to me! For the divider between the other dropdown stuff and entity previews I think it's okay to put that (and any logic around showing/hiding it) in `SearchBar`, not everything needs to be 100% configurable

cea2aj:
Another approach worth considering is only making the top divider appear if there are autocomplete results or recent searches. This would allow the code to not have to worry about the context of whether or not other things are rendered. This is the same idea as single direction padding. Each FAQ would have a divider at the top, and the entity previews would always put a divider at the top when there are 1 or more entities.

nmanu1:
> Another approach worth considering is only making the top divider (under the search input) appear if there are autocomplete results or recent searches. This would allow the code to not have to worry about the context of whether or not other things are rendered. This is the same idea as single direction padding. Each FAQ would have a divider at the top, and the entity previews would always put a divider at the top when there are 1 or more entities.

That would work, except the divider for the search input and for the entity previews seem to have slightly different styling (the length is a little longer and there's no top/bottom margin). I can try messing with the styling to make them interchangeable, but I think it might be simpler to control the entity preview dividers instead because that would give some flexibility in styling and make it so we don't depend on the entity preview rendering to separate the search input from the dropdown.

src/components/SampleVisualSearchBar.tsx
nmanu1: We do! Added it back

src/components/EntityPreviews.tsx
nmanu1: Updated!

src/components/SearchBar.tsx
```diff
@@ -310,6 +310,7 @@ export default function SearchBar({
           <StyledDropdownMenu cssClasses={cssClasses}>
             {renderRecentSearches()}
             {renderQuerySuggestions()}
+            {showEntityPreviewsDivider && <div className='h-px bg-gray-200 mt-1 mb-4 mx-3.5'></div>}
```
oshi97, line 313: can we add this to the `cssClasses` prop? not everything needs to be configurable but we also kind of have this pattern right now where all css classes are configurable

nmanu1: sure, added!

PR #68, opened 2022-02-08
Refactor StaticFilters and Facets

cea2aj:
Refactor StaticFilters and Facets to use a common Filters component

The common Filters component allows all of the styling and functionality (collapsible and searchable) to be reused. The purpose of the StaticFilters and Facets components is now to gather the filters data and click handlers and pass them to the Filters component. The Filters components renders the possible filters for various fields, and the CollapsibleLabel allows the Filters to be collapsed.

The facets have a number to indicate the count (e.g. 'Cost (3)'), and this number with the parenthesis will now be used in the Levenshtein matching since only the label is passed to the Filters component. This seems to work just fine with testing.

J=SLAP-1856
TEST=manual

Compare side by side with the hosted starter app and confirm styling is the same. Check desktop and mobile view. Enable


src/components/FilterDisplay.tsx
```diff
@@ -0,0 +1,102 @@
...
+    searchable,
+    collapsible,
+    placeholderText = 'Search here...',
+    defaultExpanded,
+    customCssClasses,
+    cssCompositionMethod
+  } = props;
+  const [ searchTerm, setSearchTerm ] = useState('');
+  const options = searchable
+    ? allOptions.filter(o => isLevenshteinMatch(o.label.toLowerCase(), searchTerm.toLowerCase()))
```
nmanu1, line 59: can we just use `isCloseMatch()` for this like we do to filter recent searches?

cea2aj: ah, I thought we had something like that but couldn't find it. I'll update!

package.json
```diff
@@ -9,6 +9,7 @@
     "@tailwindcss/forms": "^0.4.0",
     "@yext/answers-headless-react": "^1.1.0-beta.3",
     "classnames": "^2.3.1",
+    "js-levenshtein": "^1.1.6",
```
oshi97, line 12: where is this used?

src/components/Facets.tsx
```diff
@@ -1,36 +1,39 @@
...
+import { useAnswersState, useAnswersActions, Filter } from '@yext/answers-headless-react'
 import { CompositionMethod, useComposedCssClasses } from '../hooks/useComposedCssClasses';
-import Facet,{ FacetConfig, FacetCssClasses } from './Facet';
-import { Divider } from './StaticFilters';
+import Filters, { FiltersCssClasses } from './Filters';
 
-
-interface FacetsProps {
-  searchOnChange?: boolean,
+export interface FacetConfig {
```
oshi97, line 5: nit but can we have the props at the top?

package.json
cea2aj: It's not needed anymore after I updated to use `isCloseMatch` so I'll remove it

src/components/FilterDisplay.tsx
```diff
@@ -0,0 +1,102 @@
...
+              value: option.value
+            }
+            return renderCheckboxOption({
+              option: { id: `${index}`, label: option.label },
+              onClick: selected => option.onClick?.(filter, selected),
+              selected: option.isSelected,
+              cssClasses
+            });
+          })}
+        </div>
```
oshi97, line 98: it feels like most of what this component is doing is providing this toggle? is it possible to have these lines instead be specified through props.children? I think that would help divide up responsibility (no need to pass the big options array as a prop here), while also making things more flexible

cea2aj: I think props.children makes sense here. Are you thinking we pass the input element in as props children as well? Or just the rendering of the options?

src/components/Filters.tsx
```diff
@@ -0,0 +1,64 @@
...
+}
+
+interface DividerProps {
+  customCssClasses?: {
+    divider?: string
+  },
+  cssCompositionMethod?: CompositionMethod
+}
+
+export function Divider({ customCssClasses, cssCompositionMethod }: DividerProps) {
```
yen-tt, line 58: I think this divider is the same as the one in StaticFilters file. Maybe we could have this in its own file to avoid duplicate

src/components/StaticFilters.tsx
```diff
@@ -54,38 +36,32 @@ export default function StaticFilters(props: StaticFiltersProps): JSX.Element {
...
-        </fieldset>
-      })}
-    </div>
+    <Filters
+      filterConfigs={filterConfigs}
+      customCssClasses={customCssClasses}
+      cssCompositionMethod={cssCompositionMethod}
+    />
   );
 }
```
yen-tt, line 67: Forgot to remove the divider below? It's moved to Filters.tsx right?

src/components/FilterDisplay.tsx
oshi97: ooh hmm, I forgot about the input element but I think that makes sense!

src/components/Facets.tsx
cea2aj: sure

src/components/Facets.tsx
```diff
@@ -39,55 +42,71 @@ export default function Facets (props: FacetsProps): JSX.Element {
...
       const overrideConfig = facetConfigs?.[facet.fieldId] ?? {};
       const config = {
         searchable,
         collapsible,
         defaultExpanded,
         ...overrideConfig
       }
-      return (
-        <div key={facet.fieldId}>
-          <Facet
```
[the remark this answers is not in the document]
cea2aj: yep, my bad I forgot to remove

src/components/FilterDisplay.tsx
cea2aj: I refactored and and I was left with a label which supported collapsibility, so I renamed `FilterDisplay` to `CollapsibleLabel`

src/components/Filters.tsx
```diff
@@ -0,0 +1,143 @@
...
+              })}
+              <div className={cssClasses.optionsContainer}> 
+                {options.map((option, index) => {
+                  const filter = {
+                    fieldId: option.fieldId,
+                    matcher: Matcher.Equals,
+                    value: option.value
+                  }
+                  return renderCheckboxOption({
+                    option: { id: `${index}`, label: option.label },
```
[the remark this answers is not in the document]
cea2aj: I wasn't seeing any errors, but I don't think it will hurt so I updated it!

yen-tt (approved):
looks great!

PR #70, opened 2022-02-14
use Filters compound component for Facets and StaticFilters

oshi97:
J=SLAP-1883
TEST=manual

styling looks the same as before
FilterDisplayManager works
both StaticFilters and Facets work


tailwind.config.js
```diff
@@ -1,5 +1,9 @@
 module.exports = {
-  content: ['./src/**/*.{js,jsx,ts,tsx}', './public/index.html'],
+  content: [
+    './src/**/*.{js,jsx,ts,tsx}',
+    './public/index.html',
+    './node_modules/@yext/answers-react-components/lib/**/*.js'
```
nmanu1, line 5: for my own understanding, why does this need to be added? does it otherwise not get the styling from the component lib?

src/pages/EventsPage.tsx
```diff
@@ -7,49 +7,45 @@ import SpellCheck from '../components/SpellCheck';
...
   verticalKey: string
 }) {
   const { pageView } = useContext(PageViewContext);
   usePageSetupEffect(verticalKey);
 
+  function renderStaticFilters() {
+    return (
+      <Filters.StaticFilters>
+        <Filters.FilterGroup defaultFieldId='venueName'>
+          <Filters.CollapsibleLabel label='Payment Options'/>
```
nmanu1, line 26: nit: should this label be "Venue"?

tailwind.config.js
oshi97: Kind of. Tailwind optimizes the css bundle to not include unused styling. If the component lib happens to use styling that is used by the main src directory, those styles will still be present in the bundle. But if the component lib uses styling that is not present anywhere in the src dir then those styles will not be present. You can see that the checkboxes lose a lot of styling if you remove this line

src/pages/EventsPage.tsx
oshi97: yes, updated!

PR #71, opened 2022-02-28
Change usePageSetupEffect to useInitialSearch

nmanu1:
Rename `usePageSetupEffect` to `useInitialSearch`. Rework it have Headless clear the previous searcher's state and add a config for the initial search that accepts a `verticalKey`, `defaultInitialSearch`, `verticalLimit`, and/or `universalLimit`. The default initial search should only be used if there is no user-entered query (including an empty string query) from the previous searcher (i.e. there is no query param stored in the URL). If used, the default initial search query should not be stored in the URL. So, the `Navigation` component links were updated to only add the query param if there is a previous query.

Also, make a tweak to the styling for the collapse filters button on mobile to match the mocks.

J=SLAP-1910
TEST=manual

- Check that specifying universal or vertical limits displays the correct number of results
- Check that the `defaultInitialSearch` is only used if there is no previous query persisted in the URL
- Check that the `defaultInitialSearch` is not persisted in the URL
- Check that a user-entered query is used on a new searcher even if that searcher has a `defaultInitialSearch` defined
- Check that if there is no `defaultInitialSearch` or query param in the URL, an empty string query is used


src/hooks/useInitialSearch.tsx
```diff
@@ -0,0 +1,74 @@
...
+  useLayoutEffect(() => {
+    verticalKey
+      ? answersActions.setVertical(verticalKey)
+      : answersActions.setUniversal();
+
+    verticalLimit && answersActions.setVerticalLimit(verticalLimit);
+    universalLimit && answersActions.setUniversalLimit(universalLimit);
+
+    const executeQuery = async () => {
+      let searchIntents: SearchIntent[] = [];
```
yen-tt, line 39: I think this is not use anywhere else beside in the if statement below, can we let directly instantiate it in line 41?

src/components/Navigation.tsx
```diff
@@ -160,12 +165,14 @@ function renderLink(
     [cssClasses.navLinkContainer___inActive ?? '']: !isActiveLink,
     [cssClasses.navLinkContainer___active ?? '']: isActiveLink
   });
+  
+  const path = query !== null ? `${to}?query=${query}` : `${to}`;
```
oshi97, line 169: I think ``` `${to}` ``` can just be `to`

src/hooks/useInitialSearch.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+    const executeQuery = async () => {
+      if (!answersActions.state.location.userLocation) {
+        let searchIntents = await getSearchIntents(answersActions, !!verticalKey) || [];
+        await updateLocationIfNeeded(answersActions, searchIntents);
+      }
+
+      let queryParam: string | null = null;
+      if (browserLocation.search) {
+        const params = new URLSearchParams(browserLocation.search);
```
oshi97, line 46: I don't think we can use react-router's "useSearchParams" hook since that was added in v6 and we're on v6, but maybe we can pull this search param parsing logic out into our own hook? we have this logic repeated in Navigation

src/hooks/useInitialSearch.tsx
```diff
@@ -0,0 +1,73 @@
...
+   * an empty string. */
+  defaultInitialSearch?: string,
+  /** The maximum number of results to display on a vertical page. */
+  verticalLimit?: number,
+  /** The maximum number of results to display per vertical on a universal page. */
+  universalLimit?: UniversalLimit
+}
+
+/**
+ * Clears the state from the previous searcher and performs the initial search on a page.
```
oshi97, line 21: maybe we could clarify here that the initial search is also run when switching navigation tabs i.e. verticals?

src/hooks/useInitialSearch.tsx
```diff
@@ -0,0 +1,73 @@
...
+        const params = new URLSearchParams(browserLocation.search);
+        queryParam = params.get('query');
+      }
+      queryParam != null
+        ? answersActions.setQuery(queryParam)
+        : answersActions.setQuery(defaultInitialSearch);
+
+      if (browserLocation.state?.querySource) {
+        const querySource = answersActions.state.query.querySource;
+        answersActions.setQuerySource(browserLocation.state.querySource);
```
cea2aj, line 55: For my own edification, when is querySource set in the browserLocation.state?

cea2aj: ooh, I see it's set with `browserHistory.push` in the SearchBar

cea2aj: When we switch over to the `SearchBar` from the standalone library, we'll need to make sure to push the querySource to the browserHistory with the `onSelectVerticalLink` callback

nmanu1: At that point we should double check with Rose because this was added when the VA spec called for the source to be set as Autocomplete from vertical links. This was removed from the spec, so we might be able to remove this `querySource` logic and `BrowserState` entirely.

oshi97:
should this hook be moved to the standalone component lib?

src/hooks/useInitialSearch.tsx
nmanu1: sure, I made a `useQueryParam` hook

src/hooks/useInitialSearch.tsx
nmanu1: updated!

src/hooks/useQueryParam.tsx
```diff
@@ -0,0 +1,11 @@
+import { useLocation } from "react-router-dom";
+
+export function useQueryParam(): string | null {
```
oshi97, line 3: `useQueryParam` is a little confusing because the search params are also known themselves as "query params" whether they are specifically for `?query=blah` or not (we've run into this naming conflict a few times before in our pre react libraries haha). I was thinking we could just make this a general "search param" hook, and pass in `'query'` as an argument to the hook? It's annoying how generic the vocabulary is surrounding this stuff I couldn't even write this comment without using the same word to mean 2 different things

nmanu1: ah that makes sense, I'll update

nmanu1:
> should this hook be moved to the standalone component lib? kinda feels like a general thing we're going to use

I agree that this hook would probably be reused, but since it's only used at the page-level and not in any of the components in the component lib, I think it makes more sense to keep this here. I can see this being something that people might want to tailor to their app

oshi97 (approved):
pr lgtm! still curious what rose thinks about the VA spec stuff, but we can update after since that'll probably have other code changes anyways

PR #72, opened 2022-03-07
split PageViewContext into mobile/desktop and filter view contexts

yen-tt:
split PageViewContext into two contexts:

- PageViewContext will manage two views of Browser page, Desktop or Mobile, with the help of a hook library @react-hook/window-size
- FilterViewContext will manage two views of Filter components, hidden or visible.
- Move ResponsiveFilter from component lib to this repo

J=SLAP-1906
TEST=manual

shrink and expand window size, see that filter is visible when it's in desktop view (ViewFiltersButton is hidden). For mobile view, see that filter is visible when click on viewFiltersButton and is hidden otherwise.


src/components/ResponsiveDivider.tsx
```diff
@@ -0,0 +1,23 @@
...
+  /** CSS class names applied to the divider equal to or below the mobileBreakpoint. */
+  mobileClassName?: string
+};
+
+
+export function ResponsiveDivider(props: ResponsiveDividerProps): JSX.Element {
+  const {
+    desktopClassName = 'w-full h-px bg-gray-200 my-4',
+    mobileClassName = 'w-screen relative right-4 border-t border-gray-200 my-4',
+    ...divProps
```
oshi97, line 16: if we're not going to include `& ComponentPropsWithRef<'div'>` in the prop interface this `divProps` is always just an empty object, I would just remove it

src/pages/LocationsPage.tsx
```diff
@@ -29,27 +30,29 @@ const filterSearchFields = [{
...
-      <FilterDisplayManager>
-        <FilterSearch
-          label='Filter Search'
-          sectioned={true}
-          searchFields={filterSearchFields}/>
-        <Filters.ResponsiveDivider />
-        <Facets/>
-      </FilterDisplayManager>
-      { (pageView === PageView.Desktop || pageView === PageView.FiltersHiddenMobile) &&
+      <FilterViewContext.Provider value={{ filterView, setFilterView }}>
```
oshi97, line 40: Is it possible to move this FilterView provider higher up so we don't have to repeat it in every page? In my mind ideally these BlahPage.tsx files are just "here is the components you want without any fancy logic". I considered putting this in App.tsx or even adding it to routeConfig.tsx but I'm not huge fan of having it for pages that don't need filters, and even less a fan of making routeConfig complicated. Just wanted to poke your brain to see if you had any thoughts? If not I think what we have here works!

oshi97: Looks like connor made a similar comment on the pb-prototype version of this PR, putting this in App.tsx here as well sgtm!

cea2aj: I agree, it would be good to keep these pages very simple. If we go with `Visible` and `NotVisible` for FilterView, we can simply set it to `NotVisible` for pages which don't have filters. I remember there has been talk in the past about putting filters on universal pages, so moving the FilterView to the whole app would make even more sense if we implement that in the future

yen-tt: `NotVisible` and `Hidden` convey the same, so I think we can keep the naming as is? As for updating the value for pages without filters, we could add the same logic across the pages to set the value inside a useLayoutEffect call of sort, but what do you think about putting it in useInitialSearch with a new prop `filterView`?

src/pages/LocationsPage.tsx
```diff
@@ -39,35 +42,33 @@ export default function LocationsPage({ verticalKey }: {
...
+            { label: 'FAQs', verticalKey: 'faqs' },
+            { label: 'Jobs', verticalKey: 'jobs' },
+            { label: 'Events', verticalKey: 'events' }
+          ]}
+        />
+        <VerticalResults
+          CardComponent={StandardCard}
+        />
+        <LocationBias />
+      </div>}
```
oshi97, line 71: brace on next line for all these

src/context/FilterViewContext.tsx
```diff
@@ -0,0 +1,28 @@
...
+  setFilterView: () => {
+    console.warn('setFilterView is undefined');
+  }
+};
+export const FilterViewContext = createContext<FilterViewContextInterface>(filterViewContextDefault);
+
+export function FilterViewContextProvider({ children }: PropsWithChildren<{}>): JSX.Element {
+  const [filterView, setFilterView] = useState<FilterView>(FilterView.Visible);
+  return (
+    <FilterViewContext.Provider value={{ filterView, setFilterView }}>
```
oshi97, line 24: would it be good to memoify this object?

yen-tt: yup!

cea2aj (approved):

oshi97 (approved):

